PR #12, opened 2022-06-03
patched productpage-v1 and added details to igw bookinfo

AlexeySkvortsov:


AlexeySkvortsov:
pushed new changes for productpage-v1

PR #62, opened 2022-07-25
working towards multi-region, independant mp setup

smarunich:


smarunich:
not merging in favour to alpha branch

PR #110, opened 2022-08-08
More outputs

smarunich:


smarunich:
addresses https://github.com/smarunich/tetrate-service-bridge-sandbox/issues/104

PR #121, opened 2022-10-11
Update variables.tf

shamusx:
Azure removed v1.23.5
verified 1.23.12 works

verified with `az aks get-versions --location westus2`
```bash
      "default": true,
      "isPreview": null,
      "orchestratorType": "Kubernetes",
      "orchestratorVersion": "1.23.12",
```


smarunich:
@shamusx should we change just to 1.23? wondering if it is supported...

PR #125, opened 2022-10-18
Do not require cloud providers that are not used

### LINKED ISSUE #124: Only require providers that are actually used
When deploying an MP, access to cloud providers is required (AWS, Azure, Google), to retrieve the remote state and load the `register_fqn` modules.
However, we should only require the tooling to be configured for the providers being actually used. For example, when using a full GCP setup and trying to just deploy an MP to GCP, the following errors appear as it attempts to get info from AWS and Azure anyway:

```
 Error: error configuring Terraform AWS Provider: no valid credential sources for Terraform AWS Provider found.
│ 
│ Please see https://registry.terraform.io/providers/hashicorp/aws
│ for more information about providing credentials.
│ 
│ Error: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, request send failed, Get "http://169.254.169.254/latest/meta-data/iam/security-credentials/": dial tcp 169.254.169.254:80: connect: host is down
│ 
│ 
│   with provider["registry.terraform.io/hashicorp/aws"],
│   on <empty> line 0:
│   (source code not available)
│ 
╵
╷
│ Error: building account: getting authenticated object ID: parsing json result from the Azure CLI: waiting for the Azure CLI: exit status 1: ERROR: AADSTS700082: The refresh token has expired due to inactivity. The token was issued on 2022-01-13T08:57:31.0489514Z and was inactive for 90.00:00:00.
│ Trace ID: e72376a9-89be-43fc-8a74-e03f905e1d00
│ Correlation ID: 3267c560-ae12-4b5e-b5c3-350e4593422d
│ Timestamp: 2022-10-17 22:08:15Z
│ To re-authenticate, please run:
│ az login --scope https://graph.microsoft.com//.default
│ 
│   with provider["registry.terraform.io/hashicorp/azurerm"],
│   on providers.tf line 31, in provider "azurerm":
│   31: provider "azurerm" {
│ 
```

nacx:
Fixes #124 

Refactors a bit the TSB module and the usage of the `register_fqn` to avoid requiring configuration for providers that aren't used. This way when deploying, say, only to GCP you don't even need to have the Azure or AWS stuff configured locally.


Makefile
```diff
@@ -13,20 +13,18 @@ all: tsb
 
 .PHONY: help
 help : Makefile
-	@sed -n 's/^##//p' $<
+	@awk 'BEGIN {FS = ":.*##"; printf "Usage:\n"} \
+			/^[.a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36mmake %-15s\033[0m %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
```
nacx, line 17: This is a small tweak to have the output properly aligned.

Makefile
```diff
@@ -89,28 +85,28 @@ gcp_k8s: init
...
+		cloud=`jq -r '.tsb_mp.cloud' terraform.tfvars.json`; \
 		cd "tsb/mp"; \
 		terraform workspace select default; \
 		terraform init; \
 		terraform apply ${terraform_apply_args} -target=module.cert-manager -target=module.es -var-file="../../terraform.tfvars.json"; \
 		terraform apply ${terraform_apply_args} -target=module.tsb_mp.kubectl_manifest.manifests_certs -var-file="../../terraform.tfvars.json"; \
 		terraform apply ${terraform_apply_args} -var-file="../../terraform.tfvars.json"; \
-		terraform apply ${terraform_apply_args} -target=module.aws_register_fqdn -target=module.azure_register_fqdn -target=module.gcp_register_fqdn -var-file="../../terraform.tfvars.json"; \
+		terraform -chdir=../register_fqdn/$$cloud init; \
+		terraform -chdir=../register_fqdn/$$cloud apply ${terraform_apply_args} -var-file="../../../terraform.tfvars.json"; \
```
nacx: This way we can only reference the cloud we need. I'm not sure it is fully safe to the init here like this, though, my TF knowledge is quite limited.

infra/gcp/main.tf
```diff
@@ -35,6 +35,7 @@ module "gcp_jumpbox" {
   vpc_subnet              = module.gcp_base[0].vpc_subnets[0]
   tsb_version             = var.tsb_version
   jumpbox_username        = var.jumpbox_username
+  machine_type            = var.jumpbox_machine_type
```
nacx, line 38: Exposing it because the current default is not available in newer regions such as europe-southwest.

modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
```diff
@@ -19,7 +19,7 @@ spec:
       selfSigned: true
   managementPlane:
     host: ${tsb_fqdn}
-    port: 8443
+    port: 443
```
nacx, line 22: Worth changing to 443 as it looks like a better default?

modules/tsb/mp/manifests/tsb/managementplane-values.yaml.tmpl
```diff
@@ -49,11 +38,24 @@ spec:
           signingKey: private.key
       signingKeysSecret: iam-signing-key
   components:
+    frontEnvoy:
+      port: 443
+    webUI:
```
nacx, line 43: I think this was misplaced outside the `components` field.

tsb/mp/main.tf
```diff
@@ -55,34 +39,9 @@ module "tsb_mp" {
...
-
-module "azure_register_fqdn" {
-  count    = var.dns_provider == "azure" ? 1 : 0
-  source   = "../../modules/azure/register_fqdn"
-  dns_zone = "azure.cx.tetrate.info"
-  fqdn     = var.tsb_fqdn
-  address  = module.tsb_mp.ingress_ip != "" ? module.tsb_mp.ingress_ip : module.tsb_mp.ingress_hostname
-}
-
-module "gcp_register_fqdn" {
```
nacx, line 82: The register modules are now called directly from the Makefile to be able to only require the ones that match the cloud being used and do not require any additional local config for other clouds.

modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
smarunich: absolutely, please update this one as well: https://github.com/smarunich/tetrate-service-bridge-sandbox/blob/1a8e34c6a27cf7067d2688d24f3566201b1c1f6e/modules/tsb/cp/manifests/tctl/tctl-controlplane.sh.tmpl#L1 to use 443

nacx: Done

nacx:
@smarunich I've updated the PR so we don't need any additional addon module for the register the FQDN and managed to directly invoke the shared module from the Makefile.
This should be ready for review/merge.

modules/tsb/mp/manifests/tsb/managementplane-values.yaml.tmpl
smarunich: any reason behind removal of ```SESSION_AGE_IN_MINUTES``` or is it wrong?

smarunich (approved):

PR #130, opened 2022-10-22
dns updates based on @nacx https://github.com/smarunich/tetrate-service-bridge-sandbox/pull/129

smarunich:
updates based on https://github.com/smarunich/tetrate-service-bridge-sandbox/pull/129


Makefile
```diff
@@ -101,8 +101,8 @@ tsb_mp:  ## Deploys MP
 		terraform output ${terraform_output_args} | jq . > ../../outputs/terraform_outputs/terraform-tsb-mp.json; \
 		fqdn=`jq -r '.tsb_fqdn' ../../terraform.tfvars.json`; \
 		address=`jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" ../../outputs/terraform_outputs/terraform-tsb-mp.json`; \
-		terraform -chdir=../../modules/$$cloud/register_fqdn init; \
-		terraform -chdir=../../modules/$$cloud/register_fqdn apply ${terraform_apply_args} -var=address=$$address -var=fqdn=$$fqdn; \
+		terraform -chdir=../fqdn/$$cloud init; \
+		terraform -chdir=../fqdn/$$cloud apply ${terraform_apply_args} -var-file="../../../terraform.tfvars.json" -var=address=$$address -var=fqdn=$$fqdn; \
```
nacx, line 105: This module doesn't need all the variables, that's why I was not passing them. Why do we need to pass the entire file?
Also, the path looks wrong, shouldn't it be `../../terraform.tfvars.json`?

tsb/fqdn/aws/main.tf
```diff
@@ -0,0 +1,12 @@
+data "terraform_remote_state" "infra" {
+  backend = "local"
+  config = {
+    path = "../../../infra/${var.tsb_mp["cloud"]}/terraform.tfstate.d/${var.tsb_mp["cloud"]}-${var.tsb_mp["cluster_id"]}-${local.k8s_regions[tonumber(var.tsb_mp["cluster_id"])]}/terraform.tfstate"
+  }
+}
```
nacx: Why is this needed, here and in the other modules, if we don't need to access the cluster info to register the fqdn?

tsb/fqdn/aws/variables.tf
```diff
@@ -0,0 +1,18 @@
+variable "dns_zone" {
+    default = null
+}
+variable "fqdn" {
+}
+variable "address" {
+}
+
+variable "tsb_mp" {
```
nacx: Same question than before, we shouldn't be needing this. What am I missing?

Makefile
smarunich: as I have switched to the terraform_remote_state instead of approaching it with jsondecode and sub-local vars, there is no impact on the performance here, but it is a manageable consistency

tsb/fqdn/aws/main.tf
smarunich: I have added it for the sake of consistency so in the future, whoever wants to expand zones similar to gcp will follow the format instead of jsondecode, however you are right, we can take it off here.

tsb/fqdn/aws/variables.tf
smarunich: not need at this stage, so can be taken down - but more a consistency matter for the further expansion based on the infra state, @nauticalmike is prepping ocp module (to deploy ocp at gcp, azure, etc) so we may be of use... however you are right, it's solely optional and not needed at this stage

Makefile
smarunich: in regards to the path, it requires triple ```../../..``` as terraform.tfvars.json is three levels above the tsb/fqdn/$$cloud 
... 
I am doing manual tests of the PR... so far looks good, however I have not tested the private DNS

tsb/mp/main.tf
```diff
@@ -1,26 +1,25 @@
 data "terraform_remote_state" "infra" {
-  count   = length(local.k8s_regions)
   backend = "local"
   config = {
-    path = "../../infra/${var.tsb_mp["cloud"]}/terraform.tfstate.d/${var.tsb_mp["cloud"]}-${count.index}-${local.k8s_regions[count.index]}/terraform.tfstate"
+    path = "../../infra/${var.tsb_mp["cloud"]}/terraform.tfstate.d/${var.tsb_mp["cloud"]}-${var.tsb_mp["cluster_id"]}-${local.k8s_regions[tonumber(var.tsb_mp["cluster_id"])]}/terraform.tfstate"
```
smarunich, line 4: @nacx fyi, i have updated the section to the a single state reference of the target MP cluster, defined by tsb_mp variable, before it has a ```count.index``` that was pointing to the cluster_id of MP, which would be always a 0, so I switched to the single state with the reference to tsb_mp.cluster_id variable

modules/gcp/register_fqdn/main.tf
```diff
@@ -1,19 +1,75 @@
...
+  type         = "A"
+  ttl          = 300
+
+  rrdatas = [data.dns_a_record_set.tsb.addrs[0]]
+}
+
+## Private DNS Zone
+
+data "google_compute_network" "tsb" {
+  count   = local.private_zone ? 1 : 0
```
smarunich, line 47: added a "count" as it is a private zone related function @nacx jfyi

tsb/mp/main.tf
nacx: Yah, I saw that pattern and I have a local change fixing that in the infra folders as well. Since our Makefile invocations are always driven individually by the index, there is no need for the sound in an of them and we can reuse the pattern everywhere, I think. I'll submit a PR once the previous one is merged.

modules/gcp/register_fqdn/main.tf
nacx: It doesn't harm to have it without the count, as this is the network where the cluster is attached to. It always exists.

tsb/fqdn/aws/variables.tf
nacx: I think it is better to not introduce unused variables for the sake of consistency. Whilst I see what you're trying to achieve, I think it brings, now, more confusion than value. For the new reader of the repo (like me or any customer looking at it), it just makes it harder to understand, because you have to figure out what is really needed and what is just noise for the sake of developer convenience.

In my opinion, we shouldn't introduce things until they're needed. Consistency is fine, but where appropriate, and IMHO, in this cases, it makes understanding the whole stuff harder.

PR #132, opened 2022-10-22
Renamed status to state, added tf ```.terraform``` cache folder clean up and providers version lock`.terraform.lock.hcl`.

smarunich:
Renamed status to state, added tf ```.terraform``` cache folder clean up and providers version lock`.terraform.lock.hcl`.

as play is run, lock is generated for dependencies, so if the same folder is used to destroy/re-build the older cache will be utilized.

as per https://developer.hashicorp.com/terraform/cli/init


smarunich:
@kurktchiev @shamusx  jfyi

PR #142, opened 2022-10-31
Change let (index increment) to arithmetic expansion

boeboe:
"let"... the build-in bash command to increment index in the Makefile is not available on Ubuntu, so changed it to arithmetic expansion instead.

REF: https://linuxize.com/post/bash-increment-decrement-variable

```
#!/bin/sh

i=0

let i++; # this will fail on Ubuntu and works on MacOS
echo $i

i=$((i+1)); # works both on Ubuntu and MacOS
echo $i


i=$(expr $i + 1); # works both on Ubuntu and MacOS
echo $i


i=`expr $i + 1`; # works both on Ubuntu and MacOS
echo $i
```


smarunich (approved):
I remember I had issues in the past... so i switched to let, but don't remember... agree let's keep the dependencies lower... as far as you tested and did not observe issues we should be good.

PR #143, opened 2022-11-01
enable gcp, azure and aws to use dev builds

smarunich:


modules/aws/jumpbox/jumpbox.userdata
```diff
@@ -21,19 +22,41 @@ users:
...
-        --username "${tsb_image_sync_username}" \
-        --apikey "${tsb_image_sync_apikey}" \
-        --registry "${registry}" \
-        --accept-eula \
-        --parallel
+      export PROJECT_CR="${registry}"
+      # to initialize GCR and use skopeo for TSB images sync below
+      docker pull busybox
+      docker tag busybox $PROJECT_CR/busybox
+      docker push $PROJECT_CR/busybox
```
nacx, line 33: Why is this image needed?

modules/aws/jumpbox/jumpbox.userdata
```diff
@@ -21,19 +22,41 @@ users:
...
+      curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+      helm repo add tetrate-tsb-helm 'https://charts.dl.tetrate.io/public/helm/charts/'
+      helm repo update
```
nacx: Can we move the helm bits inside the development conditional? We shouldn't need it for normal tctl-based installs.

nacx (approved):
Overall LGTM. Just a couple small nits

modules/aws/jumpbox/jumpbox.userdata
smarunich: skopeo does read a metadata first to push, so as there is no images at GCR - it fails, I have to do a silly push to unblock skopeo for the sync (I don't want to do docker pull/push) I want to keep the jumpbox.userdata the same - no exceptions across clouds - it is easier to maintain that way, so the workaround is primarily for GCR.

modules/aws/jumpbox/jumpbox.userdata
smarunich: ack, done

modules/aws/jumpbox/jumpbox.userdata
nacx: I'm not sure I understand the issue. Could you detail it a bit more?
Also, given that we sync images one by one, does it happen if you use `skopeo copy` instead of `sync`?

smarunich: I have been hitting https://github.com/containers/skopeo/issues/1627 with ```skopeo copy```, the issue with ```skopeo sync``` is GCR or ECR is not being initialized (no containers uploaded), i.e. no metadata to read for skopeo sync when it starts the execution, so skopeo sync fails. Steps to reproduce:

1. Create a new GCP project
2. Try to upload using skopeo sync (the upload will fail)
3. Push the docker image towards GCR under a new GCP project
4. Re-try skopeo sync, the upload will succeed. 

I can use a newer version of skopeo container to move to copy, but I will open an issue to improve this later. Technically we may have other containers at this repo, so for now I will keep it as a workaround.

smarunich: I have opened an issue to track it: https://github.com/smarunich/tetrate-service-bridge-sandbox/issues/144

PR #145, opened 2022-11-04
[Issue23] replace Owner tag with var.owner for AWS and Azure

boeboe:
Implementing https://github.com/smarunich/tetrate-service-bridge-sandbox/issues/23

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,7 @@ destroy:  ## Destroy the environment
 		fqdn=`jq -r '.tsb_fqdn' terraform.tfvars.json`; \
 		address=`jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" outputs/terraform_outputs/terraform-tsb-mp.json`; \
 		cd "tsb/fqdn/$$cloud"; \
+		terraform init ; \
 		terraform destroy ${terraform_apply_args} -var-file="../../../terraform.tfvars.json" -var=address=$$address -var=fqdn=$$fqdn; \
 		rm -rf terraform.tfstate.d/; \
 		rm -rf terraform.tfstate; \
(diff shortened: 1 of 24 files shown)


Makefile
```diff
@@ -180,6 +180,7 @@ destroy:  ## Destroy the environment
 		fqdn=`jq -r '.tsb_fqdn' terraform.tfvars.json`; \
 		address=`jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" outputs/terraform_outputs/terraform-tsb-mp.json`; \
 		cd "tsb/fqdn/$$cloud"; \
+		terraform init ; \
```
smarunich: the state should exist for destroy... not sure why init is required here.

modules/gcp/k8s/main.tf
```diff
@@ -20,6 +20,13 @@ resource "google_container_cluster" "tsb" {
   # node pool and immediately delete it.
   remove_default_node_pool = true
   initial_node_count       = 1
+
+  resource_labels = {
+    name        = "${var.cluster_name}_tsb_sandbox_blue"
+    environment = "${var.name_prefix}_tsb"
+    owner       = var.owner
```
smarunich, line 27: so for AWS and Azure we have converged to Tetrate:Owner, but for GCP it is Owner label, can we do the same for GCP?

smarunich (changes requested):

smarunich:
overall perfect, just couple of minor comments to address

modules/gcp/k8s/main.tf
boeboe: That was my attempt, but google only allows lowercase (I had it in my first GCP commit, but had to revert that after errors)

https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements

> Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes. All characters must use UTF-8 encoding, and international characters are allowed.

smarunich: ack, thanks for the context

Makefile
smarunich: what about this one? @boeboe

boeboe: I removed it in my next commit. The reason I added is because I got `module not found` errors while testing this for all clouds with `make k8s` and and `make destroy`. I just realized that the fqdn part is ran at later stages of the `make tsb` flow.

smarunich (approved):

PR #146, opened 2022-11-07
Improve skopeo sync and automatically get internal gcr token

### LINKED ISSUE #144: Improve ```skopeo sync``` process for dev builds
As per conversation: https://github.com/smarunich/tetrate-service-bridge-sandbox/pull/143#pullrequestreview-1164080467

we are required to have a silly workaround to upload the ```busybox``` image to CR, before leveraging ```skopeo sync```

jumpbox.userdata
```
      # to initialize CR metadata and use skopeo for TSB images sync below
      docker pull busybox
      docker tag busybox $PROJECT_CR/busybox
      docker push $PROJECT_CR/busybox
```

nacx:
Fixes https://github.com/smarunich/tetrate-service-bridge-sandbox/issues/144

This changes the jumpbox sync to use `skopeo copy` and do not require to pull the random busybox image.

It also removes the need to manually pass the internal registry and token when using development versions. The internal registry is very unlikely to change, so this change encapsulates it in the `internal_registry` module instead.

The only drawback of this change is that the internal token may be generated on every run, but IMO this is a good enough tradeoff: it would only redeploy the jumpbox and at the time of re-running the `dev` version may already be pointing at a new commit (plus, how many times do we re-apply existing environments? At the time of doing that it is likely that the old token is already expired, so I think it is a good enough tradeoff for simplicity).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 # Local .terraform directories
 **/.terraform/*
 tests/*
-*.sh
 *.tfvars.json
 
 # .tfstate files
--- a/modules/aws/jumpbox/jumpbox.userdata
+++ b/modules/aws/jumpbox/jumpbox.userdata
@@ -26,26 +26,19 @@ write_files:
       curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
       install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
       ${docker_login}
-      export PROJECT_CR="${registry}"
-      # to initialize CR metadata and use skopeo for TSB images sync below
-      docker pull busybox
-      docker tag busybox $PROJECT_CR/busybox
-      docker push $PROJECT_CR/busybox
-      export TSB_VERSION="${tsb_version}"
-      if [[ "$TSB_VERSION" =~ .*"-dev" ]]; then
+      if [[ ! -z "${tetrate_internal_cr}" ]]; then
         curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
         helm repo add tetrate-tsb-helm 'https://charts.dl.tetrate.io/public/helm/charts/'
         helm repo update
         curl -Lo "/usr/local/bin/tctl" "https://binaries.dl.tetrate.io/public/raw/versions/linux-amd64-next/tctl"
         chmod +x "/usr/local/bin/tctl"
-        helm pull tetrate-tsb-helm/managementplane --version $TSB_VERSION --devel
-        tar zxvf managementplane-$TSB_VERSION.tgz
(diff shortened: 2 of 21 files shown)


.gitignore
```diff
@@ -1,7 +1,6 @@
 # Local .terraform directories
 **/.terraform/*
 tests/*
-*.sh
```
nacx, line 4: Current scripts were already excluded as they all live in `output/` or `.terraform/`, so it's safe to remove this so that modules can have `.sh` templates.

modules/aws/jumpbox/jumpbox.userdata
```diff
@@ -26,34 +26,27 @@ write_files:
...
         chmod +x "/usr/local/bin/tctl"
-        helm pull tetrate-tsb-helm/managementplane --version $TSB_VERSION --devel
-        tar zxvf managementplane-$TSB_VERSION.tgz
-        export TETRATE_INTERNAL_CR="${tetrate_internal_cr}"
-        export TETRATE_INTERNAL_CR_TOKEN="${tetrate_internal_cr_token}"
-        sed -i "s#containers.dl.tetrate.io#$TETRATE_INTERNAL_CR#g" managementplane/images.txt
-        for IMAGE in $(cat managementplane/images.txt)
+        helm pull tetrate-tsb-helm/managementplane --version ${tsb_version} --devel
+        tar zxvf managementplane-${tsb_version}.tgz
+        for IMAGE in $(cut -d/ -f2- managementplane/images.txt)
```
nacx, line 37: Minor change to avoid modifying the downloaded `images.txt`

modules/internal_registry/internal-cr-token.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+VERSION=$(jq -r '.tsb_version')
+
+if [[ "${VERSION}" =~ .*"-dev" ]]; then
+    REGISTRY="gcr.io/tetrate-internal-containers"
+    TOKEN=$(gcloud auth print-access-token)
+fi
```
nacx, line 10: Default to empty values to not require `gcloud` if it is not a dev version, and also to allow decoupling the registry from the version in other modules. With this other modules can just use `if an internal registry has been set, use it`, without having to parse the TSB version.

smarunich (approved):

modules/internal_registry/internal-cr-token.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+VERSION=$(jq -r '.tsb_version')
+
+if [[ "${VERSION}" =~ .*"-dev" ]]; then
+    REGISTRY="gcr.io/tetrate-internal-containers"
```
smarunich, line 8: I did not want to expose it on a public repo, as I am not owning the repo, but it simplifies the code so appreciated.

PR #147, opened 2022-11-08
Expose a custom helm repo for TSB install

smarunich:
The PR enables the user to:
* Expose a custom helm repo for TSB install, to include perform a dev install (requires jumpbox to be aware of the custom repo as well)
* Uplifts cert-manager to 1.9.1

### DIFF
--- a/infra/aws/main.tf
+++ b/infra/aws/main.tf
@@ -20,6 +20,7 @@ module "aws_jumpbox" {
   vpc_subnet                = module.aws_base[0].vpc_subnets[0]
   cidr                      = module.aws_base[0].cidr
   tsb_version               = var.tsb_version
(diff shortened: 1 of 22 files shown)


infra/aws/main.tf
```diff
@@ -20,6 +20,7 @@ module "aws_jumpbox" {
   vpc_subnet                = module.aws_base[0].vpc_subnets[0]
   cidr                      = module.aws_base[0].cidr
   tsb_version               = var.tsb_version
+  tsb_helm_repository       = var.tsb_helm_repository
```
nacx, line 23: Can you please update the `DEVELOPMENT_BUILD.md` accordingly?

smarunich: ah you are right, good point!

PR #154, opened 2022-11-15
custom iam signing key

smarunich:
between keeping the custom iam signing key or switching to defaults

### DIFF
--- a/modules/tsb/mp/manifests/tsb/managementplane-values.yaml.tmpl
+++ b/modules/tsb/mp/manifests/tsb/managementplane-values.yaml.tmpl
@@ -35,7 +35,7 @@ spec:
       issuers:
         - name: https://jwt.tetrate.io
           signingKey: private.key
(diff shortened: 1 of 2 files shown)


modules/tsb/mp/manifests/tsb/managementplane-values.yaml.tmpl
```diff
@@ -35,7 +35,7 @@ spec:
       issuers:
         - name: https://jwt.tetrate.io
           signingKey: private.key
-      signingKeysSecret: iam-signing-key
+      signingKeysSecret: custom-iam-signing-key
```
nacx: Probably better to remove the issuers list to use the defaults? Less internal details exposed as config.

smarunich: agree, will clean up

nacx (approved):

PR #156, opened 2022-11-16
eshop ingress with rate limiting and optional WAF

nacx:
Opening as draft because in the latest master some of the WAF features referenced here are not yet available. This depends on a couple PRs being merged to capture the last WAF WASM image with the filesystem aliases.

### DIFF
--- a/apps/eshop/README.md
+++ b/apps/eshop/README.md
@@ -11,7 +11,8 @@ It consists of two applications:
 The `payments` service is configured to introduce a 200ms latency, and the `checkout`
 service is configured to fail 20% of the requests.  
 The `eshop` ingress gateway is also
-configured to rate-limit to a max of 3 requests per second per unique client address.   
+configured to rate-limit to a max of 3 requests per second per unique client address, and with
+WAF settings that load the Core Rule Set and have request body inspection enabled. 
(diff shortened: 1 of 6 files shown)


apps/eshop/README.md
```diff
@@ -11,7 +11,8 @@ It consists of two applications:
 The `payments` service is configured to introduce a 200ms latency, and the `checkout`
 service is configured to fail 20% of the requests.  
 The `eshop` ingress gateway is also
-configured to rate-limit to a max of 3 requests per second per unique client address.   
+configured to rate-limit to a max of 3 requests per second per unique client address, and with
```
smarunich, line 14: for rate-limit ... we need to add redis (https://github.com/smarunich/tetrate-service-bridge-sandbox/issues/58), also need to update overlay (https://docs.tetrate.io/service-bridge/1.5.x/en-us/howto/rate_limiting/internal_rate_limiting), to be fair we can make it as a default installation option as it is a POC  environment...

nacx: Oh, I see. I'm too used to the demo profile :)

Anyway, I'm planning on moving all the eshop stuff to gitops folder, because that will be much easier to manage and to deploy in the desired clusters, instead of everywhere, so there are chances I'll close this PR and replace with that.

PR #157, opened 2022-11-21
istio-proxy readiness before app

shamusx:


smarunich (approved):

smarunich:
wdyt @nacx it is a second iteration of it (of the customer in a scale require the feature...)

nacx:
I'm not super familiar with that and the impact it has, but LGTM overall (this is mostly to build PoC environments).
Is this overlay commonly used enough so it would make sense to promote it to a proper setting in the `ControlPlane` CRD?

nacx (approved):

smarunich:
@nacx I have raised this one https://github.com/tetrateio/tetrate/issues/14334, trying to follow that path as you suggested.

PR #159, opened 2022-11-22
added displayName to dev and stg tenants

AlexeySkvortsov:
we need to update repo to sync displayName for tenants via gitops/argocd

### DIFF
--- a/gitops/bookinfo-direct/tsb-bookinfo.yaml
+++ b/gitops/bookinfo-direct/tsb-bookinfo.yaml
@@ -7,6 +7,7 @@ metadata:
     tsb.tetrate.io/organization: tetrate
 spec:
   description: stg
+  displayName: argocd
 ---
 apiVersion: tsb.tetrate.io/v2
 kind: Workspace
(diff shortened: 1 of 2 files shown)


gitops/bookinfo-direct/tsb-bookinfo.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     tsb.tetrate.io/organization: tetrate
 spec:
   description: stg
+  displayName: argocd
```
shamusx: displayName and name should match

gitops/bookinfo/tsb-bookinfo.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     tsb.tetrate.io/organization: tetrate
 spec:
   description: dev
+  displayName: argocd
```
shamusx: displayName and name should match

AlexeySkvortsov:
@shamusx corrected

smarunich (approved):

PR #164, opened 2022-12-03
adding support for cr-pull-secret for wasm

smarunich:
adds `cr-pull-secret` under istio-system namespace for the further use by `WASM` extensions, watch out for ECR as the `cr-pull-secret` has to be refreshed every 12 hrs, where it is a static construct for Azure and Google.


smarunich:
addresses https://github.com/smarunich/tetrate-service-bridge-sandbox/issues/163

shamusx (approved):

PR #182, opened 2022-12-21
Move ArgoCD apps out of the generic module

nacx:
This makes the ArgoCD generic module easier to reuse as it doesn't hardcode the Applications to install, but instead provides a variable to set them. The apps have been moved to the outer argocd module.


modules/addons/elastic/manifests/es.yaml
```diff
@@ -40,7 +40,7 @@ spec:
   http:
     service:
       spec:
-        type: LoadBalancer
+        type: ClusterIP
```
nacx, line 43: This is a small cleanup. I highly doubt anyone would be accessing Kibana directly to see what's there in Elastic/Skywalking. When that is done (quite uncommon) it's not a big deal to forward the port, and we save one Elastic IP.

smarunich: makes sense!

smarunich (approved):

PR #183, opened 2022-12-23
Allow to opt-out of preemptible nodes in GKE

nacx:
When building scenarios to do customer demos, having preemptible nodes adds too much uncertainty to the availability of the cluster. We still default to preemptible, but with this change we allow configuring standard VMs to create more stable environments.

cc @gonzaloserrano

### DIFF
--- a/DEVELOPMENT_BUILD.md
+++ b/DEVELOPMENT_BUILD.md
@@ -1,4 +1,4 @@
-# DEVELOPMENT_BUILD.md
+# Provisioning development versions
 
 If you want to provision the latest master build you have to:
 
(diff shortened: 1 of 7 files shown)


gonzaloserrano (approved):

smarunich:
@nacx @gonzaloserrano thank you for taking a look into this, i honestly did keep it as my "chaos" monkey approach when `testing` TSB and basically got to ignore it at this stage. I think we should change defaults to `false`, i.e. team does use the repo more to demo, including myself... not to discover and do random TSB testing by default. WDYT?

nacx:
Since no one has complained about this yet, I think it's fine to keep the default to true, as it's cheaper? :P If someone complains we can always change the default, that's easy.

nacx:
Thinking a bit more about it... I've changed the default to standard VMs, a you suggested @smarunich. Seems more reasonable. We can still set the `preemptible_nodes: true` in the tfvars json file to use spot vms., when desired.

gonzaloserrano (approved):

PR #186, opened 2022-12-23
Add rate limiting backend

### LINKED ISSUE #58: add ```redis```, rate limiters example

nacx:
Fixes https://github.com/smarunich/tetrate-service-bridge-sandbox/issues/58

Adds the rate-limit backend tot he CPs and configures the eshop with WAF and rate limiting enabled by default.


smarunich:
do you mind replicating `redis` install using `cert-manager` as an example model (to basically follow the addons model) and leveraging https://artifacthub.io/packages/helm/bitnami/redis helm chart instead - thank you!!

nacx:
One of the reason I didn't go the Helm route is because we use our very own redis instance (the envoy ratelimit-redis). I don't know to what extent is just a vanilla redis or it has other stuff, so I opted to not use an official vanilla redis Helm chart, to avoid any confusion/gotchas. WDYT?

nacx:
Moving to draft as I have a couple changes to move it to the addons folder, although I'll keep the deployment files instead of using a Helm chart, as we can't guarantee hat helm chart matches the image we're using, and there could be mismatches in startup flags, etc.

smarunich:
ack @nacx , meanwhile I will investigate further the image story, i got your point, sounds like a fair statement, also need to make an option to have the rate limit as option to choose, but enabled by default.

nacx:
The image is based on this one: https://github.com/envoyproxy/ratelimit
I think it's safer to not make assumptions on what Helm charts it can be installed with and just have a deployment+service.

nacx:
Done, the ratelimit component is now an optional addon, enabled by default.

tsb/cp/providers.tf
```diff
@@ -8,22 +8,22 @@ terraform {
 }
 
 provider "kubernetes" {
-  host                   = local.infra[var.tsb_mp["cloud"]][var.tsb_mp["cluster_id"]]["outputs"].host
-  cluster_ca_certificate = base64decode(local.infra[var.tsb_mp["cloud"]][var.tsb_mp["cluster_id"]]["outputs"].cluster_ca_certificate)
-  token                  = local.infra[var.tsb_mp["cloud"]][var.tsb_mp["cluster_id"]]["outputs"].token
+  host                   = local.infra["outputs"].host
+  cluster_ca_certificate = base64decode(local.infra["outputs"].cluster_ca_certificate)
+  token                  = local.infra["outputs"].token
```
nacx, line 13: @smarunich PTAL, I think this setup was fundamentally wrong, as we're deploying CPs here but the variables referenced the cluster_id referenced from the `mp` settings... I may be missing something, but... how could this be working when more than 1 cluster was deployed? What am I missing?

In any case I think the change is correct and the current configuration is the right one, just reading the current cluster, not mp related stuff.

modules/addons/ratelimit/main.tf
```diff
@@ -0,0 +1,41 @@
+provider "kubectl" {
+  host                   = var.k8s_host
+  cluster_ca_certificate = base64decode(var.k8s_cluster_ca_certificate)
+  token                  = var.k8s_client_token
+  load_config_file       = false
+}
+
+data "external" "redis_image" {
```
nacx: Using this to fetch the image list and always set the right version of the redis image for the configured TSB version so we don't need to hardcode it and keep updating it.

gitops/eshop/tsb-config.yaml
```diff
@@ -116,24 +116,24 @@ items:
...
+                  unit: SECOND
+      waf:
+        rules:
+          - Include @recommended-conf         # Recommended basic setup
+          - SecRuleEngine On                  # Override DetectionOnly mode in @recommended-conf
+          - SecResponseBodyAccess Off         # Override to not process response bodies
+          - Include @crs-setup-conf           # Initialize the Core Rule Set
+          - Include @owasp_crs/*.conf         # Load the Core Rule Set
+          # Custom rule to deny requests with query parameter 'id=0'
+          - SecRule ARGS:id "@eq 0" "id:1, phase:1,deny, status:403,msg:'Invalid id',log,auditlog"
```
smarunich: can we also add 
```
  - SecRule REQUEST_HEADERS:user-agent "@contains curl" "id:2, pass, log,
    logdata:'someone used curl to access'"
```
also something i want to add it is a `json` payload sanitizer using `waf`...  i.e. just to outline the functionality and ability to tamper the body.

modules/addons/ratelimit/redis-image.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+eval "$(jq -r '@sh "TSB_VERSION=\(.tsb_version) REGISTRY=\(.registry)"')"
+
+TMP_DIR=$(mktemp -d)
+helm pull tetrate-tsb-helm/controlplane --version "${TSB_VERSION}" --untar --untardir "${TMP_DIR}"
```
smarunich: it implies that... the end user has a `helm` repository added with the exactly the name... should we instead do `rest api` call towards the `cloudsmith` to retrieve the image? for example i was using this url recently `'[https://api.cloudsmith.io:443/v1/packages/tetrate/getistio-containers/?page=1&page_size=30](https://api.cloudsmith.io/v1/packages/tetrate/getistio-containers/?page=1&page_size=30)'` to get the TID images... I can work towards `TSB` example basically to use latest redis from `https://cloudsmith.io/~tetrate/repos/tsb/packages/?q=redis` output...

another lazy one can be a `curl` of `https://charts.dl.tetrate.io/public/helm/charts/controlplane-1.5.5.tgz` a helm chart? 

the best approach will be to limit the local utilities dependencies or do rest api call within terraform... i.e. helm local setup can be broken, use different names, etc...  or not having a helm at all.

modules/addons/ratelimit/manifests/ratelimit-redis.yaml.tmpl
```diff
@@ -0,0 +1,70 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ratelimit
+---
+apiVersion: apps/v1
+kind: Deployment
```
smarunich: I still have this strong feeling we can use `tsb` image and just install it using `helm` (https://artifacthub.io/packages/helm/bitnami/redis)  it will be easier to keep up with the `deployment` changes instead of `having a tsb specific` way, practically if the customer is looking to deploy rate limiters they should be using their managed copy of `redis`... so by having an external redis as part of the repo...but matching the version 

technically we should not maintain own fork or "how to do redis", but having an external redis as part of this repo, will look more like an deployment pattern, in the same way as we do use external cert-manager. WDYT?

gitops/eshop/tsb-config.yaml
nacx: There is a traffic generator installed for the eshop and this will introduce a lot of noise, because the traffic generator uses curl.. Processing the user-agent is already covered in an example in the eshop README, so I'd rather not add this rule.

modules/addons/ratelimit/manifests/ratelimit-redis.yaml.tmpl
nacx: Done

modules/addons/grafana/main.tf
```diff
@@ -13,6 +13,10 @@ provider "kubectl" {
   load_config_file       = false
 }
 
+resource "random_password" "grafana" {
```
smarunich, line 16: let's use  `"tsb_password"` as a variable to set set password for all the related tools, for example argocd password is using `tsb_password` as well.

modules/addons/ratelimit/main.tf
```diff
@@ -0,0 +1,37 @@
...
+}
+
+provider "kubectl" {
+  host                   = var.k8s_host
+  cluster_ca_certificate = base64decode(var.k8s_cluster_ca_certificate)
+  token                  = var.k8s_client_token
+  load_config_file       = false
+}
+
+resource "random_password" "redis" {
```
smarunich, line 16: let's use "tsb_password" as a variable to set set password for all the related tools, for example argocd password is using tsb_password as well.

tsb/cp/providers.tf
smarunich: indeed, it could not be working in the past, in fact the tsb_cp module k8s provider being initialized as part of tsb_cp module itself, the outer `providers.tf` is used for k8s operations outside of the modules which we don't have any yet at this stage, so it is a more ready to go placeholder at this stage.

addons/monitoring/main.tf
```diff
@@ -21,7 +21,6 @@ module "grafana" {
   k8s_cluster_ca_certificate = data.terraform_remote_state.infra.outputs.cluster_ca_certificate
   k8s_client_token           = data.terraform_remote_state.infra.outputs.token
   namespace                  = "tsb-monitoring"
-  admin_password             = var.tsb_password
```
smarunich, line 24: hopefully this is not the issue to use `var.tsb_password` it will just keep the UX more streamlined.

gitops/eshop/README.md
```diff
@@ -24,3 +22,33 @@ by configuring the environment variables as explained in the table at the end.
...
+curl -i -X POST -H "Host: eshop.tetrate.io" \
+    --data "1%27%20ORDER%20BY%203--%2B" -H "X-B3-Sampled: 1" \
+    http://${ESHOP_GW}/proxy/orders
+
+# Vulnerability scanner detection
+curl -i -H "Host: eshop.tetrate.io" -H "X-B3-Sampled: 1" \
+    --user-agent "Grabber/0.1 (X11; U; Linux i686; en-US; rv:1.7)" \
+    http://${ESHOP_GW}/proxy/orders
+
+# Custom rule that rejects an invalid ID
```
smarunich, line 51: thank you for examples!

gitops/eshop/tsb-config.yaml
```diff
@@ -116,24 +116,25 @@ items:
...
+      waf:
+        rules:
+          - Include @recommended-conf         # Recommended basic setup
+          - SecRuleEngine On                  # Override DetectionOnly mode in @recommended-conf
+          - SecResponseBodyAccess Off         # Override to not process response bodies
+          - Include @crs-setup-conf           # Initialize the Core Rule Set
+          - Include @owasp_crs/*.conf         # Load the Core Rule Set
+          # Custom rule to deny requests with query parameter 'id=0'
+          - SecRule ARGS:id "@eq 0" "id:1,phase:1,deny,status:403,msg:'Invalid id',log,auditlog"
+          - SecRule REQUEST_HEADERS:user-agent "@contains curl" "id:2,pass,log,logdata:'someone used curl to access'"
```
smarunich, line 137: I agree with your previous comments we just can keep the line commented out 
```- SecRule REQUEST_HEADERS:user-agent "@contains curl" "id:2,pass,log,logdata:'someone used curl to access'"```

File: modules/addons/ratelimit/main.tf [hunk 11]
```diff
@@ -0,0 +1,37 @@
...
+  length = 16
+}
+
+resource "helm_release" "redis" {
```
smarunich, line 20: should we set the `image.tag` to `7.0.7` to basically enforce the version rotation? like we do for `cert-manager` and `elastic/eck`, i am not worried about `grafana`, `argocd`, `prometheus` - however the `eck`, `cert-manager` and `redis` may impact the product functionality directly. wdyt?

modules/addons/ratelimit/main.tf
smarunich: ignore my previous comment on redis, agree to have a random password set for backend `redis`. It only makes sense to have a user facing tools password to be set to `tsb_password`

modules/addons/ratelimit/main.tf
```diff
@@ -0,0 +1,37 @@
...
+  count            = var.enabled ? 1 : 0
+  name             = "redis"
+  repository       = "https://charts.bitnami.com/bitnami"
+  chart            = "redis"
+  create_namespace = true
+  namespace        = "ratelimit"
```
smarunich: perhaps `tsb-ratelimit` to follow `tsb-monitoring` for namespace? or `redis` to follow `cert-manager`? i do like `tsb-ratelimit` and `tsb-monitoring`.

gitops/eshop/tsb-config.yaml
smarunich: yeah.. do forget about the noise, agree, lets comment out the rule.

addons/README.md
```diff
@@ -31,7 +31,8 @@ make monitoring
 ```
 
 `grafana` is exposed using ClusterIP type. It can be accessed by port-forwarding port 3000 to the `grafana` pod
-in the `tsb-monitoring` namespace. The username is `admin` and password is variable `tsb_password` from `tfvar`.
+in the `tsb-monitoring` namespace. The username is `admin` and the password can be found in the
+`outputs/terraform_outputs/terraform-monitoring.json` file.
```
smarunich: let's keep the password as `tsb_password` to keep the UX more streamlined.

Makefile
```diff
@@ -161,6 +161,7 @@ monitoring:  ## Deploys the TSB monitoring stack
 		terraform workspace select default; \
 		terraform init; \
 		terraform apply ${terraform_apply_args} -var-file="../../terraform.tfvars.json"; \
+		terraform output ${terraform_output_args} | jq . > ../../outputs/terraform_outputs/terraform-monitoring.json; \
```
smarunich, line 165: thank you for streaming the output to json, it does help in general to troubleshoot the envs after on too.

gitops/eshop/tsb-config.yaml
nacx: I'd rather not add this rule. There is already one rule showing how to add custom rules, which is the objective. Why this rule and not any other that shows any other SecLang keyword? Let's keep the example focused on the CRS and showing the syntax.

modules/addons/ratelimit/main.tf
nacx: I don't know what versions of redis this chart is compatible with, so I'd rather just keep it to the chart's default version.

smarunich: ack, it does push 7.0.7,  looked up over here: https://artifacthub.io/packages/helm/bitnami/redis

gitops/eshop/tsb-config.yaml
smarunich: agree

modules/addons/ratelimit/main.tf
nacx: In the Bitnami Redis chart, the version of the chart is disconnected from the Redis version. Keeping a fixed Redis image version here forces us to maintain the right version of the chart and make sure which ones are compatible. This is one of the reasons I initially wanted to use the custom deployment: depending on a non-official chart adds this extra complexity we have to deal with now. Honestly, I think it's not worth it.

modules/addons/ratelimit/main.tf
nacx: Done

modules/addons/grafana/main.tf
nacx: Ok, here and in the other user-facing things I've changed it a bit to use the `tsb_password` if set, but to generate a random secure password if not set. The `tsb_password` is no longer mandatory, and if omitted a random secure pass will be set for each component. If set, TSB, Grafana and ArgoCD will use the same tsb password.

addons/monitoring/dashboards/t1-gateways.json
```diff
@@ -1,1403 +0,0 @@
-{
```
nacx, line 1: This dashboard should have never been here; it is one for the hosted env that is not a TSB one.

modules/addons/grafana/main.tf
smarunich: even better! thank you!

smarunich (approved):

PR #191, opened 2023-01-03
adding tcp-echo placeholder

smarunich:


nacx:
LGTM.

BTW, the [obstester app](https://github.com/tetrateio/tetrate/tree/master/test/services/obs-tester) already has an echo server running on port 3000, so technically the gitops eshop app is already exposing a TCP server in each service, in case you wanna leverage that instead of pulling a new app.

smarunich:
oh cool, I will re-base to obstester! thank you @nacx

nacx (approved):

PR #200, opened 2023-01-09
Do not use `:` in Redis passwords

nacx:
If password contains the `:` character it will be read as Base64 auth and split into username and password, leading to error like this in the `ratelimit-server`:

```
time="2023-01-09T17:31:24Z" level=warning msg="enabling authentication to redis on redis-master.tsb-ratelimit.svc.cluster.local:6379 with user i>chwmz"
creating redis connection error : WRONGPASS invalid username-password pair or user is disabled.
```

### DIFF
--- a/modules/addons/ratelimit/main.tf
+++ b/modules/addons/ratelimit/main.tf
@@ -15,6 +15,8 @@ provider "kubectl" {
 
 resource "random_password" "redis" {
   length = 16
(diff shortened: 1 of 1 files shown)


smarunich:
thank you, as indeed I did a face a single time failure, but was not on a different to troubleshoot.

PR #209, opened 2023-01-12
set OAP SW_CORE_PERSISTENT_PERIOD to 5

smarunich:
as per 

```
Each OAP (MP and CP) has a period in which it "collects and flushes metrics"  in bulk.  This is to optimize performance and bandwidth 
By default it is 25s.  This is why there is such a delay in TSB UI.  However, for a demo env I drop that down to 5s so that metrics and topology changes appear nearly instantaneous
```

should we @nacx @gonzaloserrano @shamusx ?

### DIFF
--- a/modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
+++ b/modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
@@ -66,9 +66,6 @@ spec:
       certManager:
         managed: EXTERNAL
     oap:
-        oap:
-          streamingLogEnabled: true
-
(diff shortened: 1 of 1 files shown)


nacx:
I think it is more reasonable to just keep the skywalking default as it's not a big deal and this will cause much more load on elastic (this is used as the flush time to elastic as well), so probably what we wanna do is keep the defaults and expose it as a var, if someone feels strongly about changing it?
Context is here: https://tetrate.slack.com/archives/CF9RHFXNZ/p1673259999223379

smarunich:
thank you for the context @nacx

PR #210, opened 2023-01-12
Automate grafana tunnel

gonzaloserrano:
Just a small toy created while trying to understand the codebase better.

### DIFF
--- a/modules/addons/grafana/main.tf
+++ b/modules/addons/grafana/main.tf
@@ -28,6 +28,7 @@ resource "helm_release" "grafana" {
   values = [templatefile("${path.module}/manifests/grafana-values.yaml.tmpl", {
     admin_password = coalesce(var.password, random_password.grafana.result)
     service_type   = var.service_type
+    service_port   = var.service_port
   })]
 }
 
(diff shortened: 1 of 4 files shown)


modules/addons/grafana/main.tf
```diff
@@ -47,3 +48,9 @@ resource "kubectl_manifest" "tsb_dashboards" {
   for_each   = var.dashboards
   yaml_body  = data.kubectl_path_documents.tsb_dashboards[each.key].documents[0]
 }
+
+resource "local_file" "ssh_tunnel" {
```
gonzaloserrano, line 52: self-review: now that I think this should be called port-forward instead of ssh-tunnel 🤔

nacx:
JFYI in case you haven't seen it, you can set these two values in your root `.tfvars` file:
```json
"grafana_service_type": "LoadBalancer",
"argocd_service_type": "LoadBalancer",
```
and those services will have a public IP allocated so that you don't need to port-forward.

smarunich:
Please lets limit the number of custom tools/overlays such as and lets use the service type instead as @nacx mentioned, for the scripts like that lets not overload the tf.

Basically the repo is to be exposed to the customer and my chapter is to keep it as simple as possible, so lets keep the deployment patterns as production ones and limit the extras that can be addressed by the existing functionality. 

I do appreciate though having the PRs like that to be opened, as it gives a chance to spread the knowledge and intentions behind the repo.

smarunich (changes requested):
--

gonzaloserrano:
Fair enough! 👍

PR #214, opened 2023-01-16
pvc cleanup

### LINKED ISSUE #128: GCP - Project where MP deployed doesn't delete
If MP is deployed on GCP cluster - GCP project will fail to delete due to PVC generated for storage of Postgres and Elastic still existing which causes compute API disable to fail.

shamusx:
script to clean up PVCs when MP is in GCP

### DIFF
--- a/modules/gcp/jumpbox/main.tf
+++ b/modules/gcp/jumpbox/main.tf
@@ -81,6 +81,7 @@ resource "google_compute_instance" "jumpbox" {
   }
 }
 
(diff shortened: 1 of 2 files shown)


modules/gcp/jumpbox/main.tf
```diff
@@ -81,6 +81,20 @@ resource "google_compute_instance" "jumpbox" {
   }
 }
 
+resource "null_resource" "gcp_cleanup" {
```
gonzaloserrano, line 85: nit: maybe add some docs on why this is needed?

smarunich: added a comment, @gonzaloserrano

smarunich (approved):

gonzaloserrano (approved):

PR #215, opened 2023-01-18
adding more ports to enable VM onboarding demo

smarunich:
as requested by @kurktchiev adding more ports to enable VM onboarding demo


kurktchiev (approved):
lgtm. this is needed for jumpboxes to be used as VM onboards for demos

Issue #184, opened 2022-12-23
Configure mesh expansion with the repository and onboarding plane endpoint


nacx:
Today we generate an empty mesh expansion configuration in the CPs.
We could easily add some config bits to allow configuring which clusters should have mesh expansion enabled, and we could also generate a proper FQDN for the vmgateway (something like `vm<cluster_id>-<tsb_fqdn>`), configure it in the provider DNS, and generate the corresponding certificate, so that the cluster is ready to be used to onboard VMs.

I have a WIP branch for this and plan to submit a PR soon.

smarunich:
This is perfect, thank you @nacx - as we proceed we need to refactor the variables so we can customize each cluster when we define it, but probably next month or so resolution. I was thinking about the above further, and did align myself towards enabling `external-dns` and allocating a zone per cluster... so eventually the elements like that can be completely done using `gitops`... and minimize the `tf` influence, but it is a time investment on itself, basically need help or time on getting `external-dns` added for the clusters. wdyt?

nacx:
I agree. I thought several times about adding external-dns, because it would also help expose people's declared IngressGateways. I already have an external-dns setup for GCP, but lacking AWS/Azure config expertise to add it in those so I parked that :) But happy to help there. I can probably get started with the GCP bits, put the structure, etc, and then let you or someone else parameterize the bits/values for Azure/AWS

smarunich:
I can take on expanding towards AWS/Azure - actually I do know how to do it on Azure but never done it for GCP.

nacx:
JFTR, GKE example bits for external-dns: https://github.com/tetratelabs/zta-demo-2022/tree/master/k8s/gke

nacx:
Once https://github.com/tetrateio/tetrate-service-bridge-sandbox/pull/247 this could be easily achieved by configuring this overlay in the CP:
https://github.com/tetrateio/eshop-demo/blob/main/vm/02-control-plane-patch.yaml#L9-L25

PR #216, opened 2023-01-20
Add External DNS add-on

nacx:
Adds an ExternalDNS add-on to automatically configure the DNS names for applications exposed in TSB Ingress gateways. It is default configured to watch the Istio Gateways so that ingress gateways created by TSB will be automatically configured in the DNS. Proper filters have been configured so that only hostnames exposed in the same DNS zone than the MP DNS-name will be generated, to avoid generating DNS records for internal hostnames, etc.

Currently, it only adds it for GCP, but the generic add-on module is already prepared to accommodate the other cloud providers. The only thing that needs to be added to the generic add-on is the provider-specific values file and the provider-specific variables.

Since external-dns requires credentials to update the DNS records and that's different in each cloud provider, I've built it as follows:

* The generic `modules/addons/external-dns` module will accept the variables for all providers as optional, sot aht each instance can set what's needed.
* There is a module for each cloud provider in the `addons/external-dns` folder so that we can instantiate the generic module with the right values for each provider.
* Inside the provider-specific module, there is a `credentials` module to create the credentials to manage DNS records in the cloud provider. This is a separate module because it should run only once to create the service account / whatever in the provider and store the key locally in the TF outputs. Once that's done, external-dns will be configured in each cluster with those credentials.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,20 +12,13 @@ terraform_output_args = -json
 .PHONY: all
 all: tsb
 
-.PHONY: help
-help: Makefile ## This help
-	@awk 'BEGIN {FS = ":.*##"; printf "Usage:\n"} \
-			/^[.a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36mmake %-15s\033[0m %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
-
-.PHONY: init
-init:  ## Terraform init
-	@echo "Please refer to the latest instructions and terraform.tfvars.json file format at https://github.com/tetrateio/tetrate-service-bridge-sandbox#usage"
+##@ Infrastructure provisioning
 
 .PHONY: k8s
 k8s: azure_k8s aws_k8s gcp_k8s  ## Deploys k8s cluster for MP and N-number of CPs(*) 
 
 .PHONY: azure_k8s
-azure_k8s: init  ## Deploys azure k8s cluster for MP and N-number of CPs(*) leveraging AKS
+azure_k8s:  ## Deploys azure k8s cluster for MP and N-number of CPs(*) leveraging AKS
 	@/bin/sh -c '\
 		index=0; \
 		name_prefix=`jq -r '.name_prefix' terraform.tfvars.json`; \
@@ -46,7 +39,7 @@ azure_k8s: init  ## Deploys azure k8s cluster for MP and N-number of CPs(*) leve
 		'
 
 .PHONY: aws_k8s
-aws_k8s: init  ## Deploys EKS K8s cluster (CPs only)
+aws_k8s:  ## Deploys EKS K8s cluster (CPs only)
 	@/bin/sh -c '\
 		index=0; \
 		name_prefix=`jq -r '.name_prefix' terraform.tfvars.json`; \
@@ -66,7 +59,7 @@ aws_k8s: init  ## Deploys EKS K8s cluster (CPs only)
 		'
 
 .PHONY: gcp_k8s
-gcp_k8s: init  ## Deploys GKE K8s cluster (CPs only)
+gcp_k8s:  ## Deploys GKE K8s cluster (CPs only)
 	@/bin/sh -c '\
 		index=0; \
 		name_prefix=`jq -r '.name_prefix' terraform.tfvars.json`; \
@@ -86,8 +79,14 @@ gcp_k8s: init  ## Deploys GKE K8s cluster (CPs only)
 		done; \
 		'
 
+##@ TSB installation
+
+.PHONY: tsb
+tsb: k8s tsb_mp tsb_cp  ## Deploys a full environment (Infra+MP+CP)
+	@echo "Magic is on the way..."
+
 .PHONY: tsb_mp
-tsb_mp:  ## Deploys MP
+tsb_mp:  ## Deploys the TSB Management Plane
 	@echo "Refreshing k8s access tokens..."
 	@$(MAKE) k8s
 	@echo "Deploying TSB Management Plane..."
@@ -103,7 +102,6 @@ tsb_mp:  ## Deploys MP
 		fqdn=`jq -r '.tsb_fqdn' ../../terraform.tfvars.json`; \
 		address=`jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" ../../outputs/terraform_outputs/terraform-tsb-mp.json`; \
 		terraform -chdir=../fqdn/$$cloud init; \
-		terraform -chdir=../fqdn/$$cloud apply ${terraform_apply_args} -var-file="../../../terraform.tfvars.json" -var=address=$$address -var=fqdn=$$fqdn; \
 		terraform workspace select default; \
 		cd "../.."; \
 		'
@@ -128,9 +126,21 @@ tsb_cp_%:
 		done; \
 		'
 
-.PHONY: tsb
-tsb: k8s tsb_mp tsb_cp  ## Deploys a full environment (MP+CP)
-	@echo "Magic is on the way..."
+##@ Add-on installation
+
+.PHONY: monitoring
+monitoring:  ## Deploys the TSB monitoring stack
+	@echo "Deploying TSB monitoring stack..."
+	@$(MAKE) k8s
+	@/bin/sh -c '\
+		cd "addons/monitoring"; \
+		terraform workspace select default; \
+		terraform init; \
+		terraform apply ${terraform_apply_args} -var-file="../../terraform.tfvars.json"; \
+		terraform output ${terraform_output_args} | jq . > ../../outputs/terraform_outputs/terraform-monitoring.json; \
+		terraform workspace select default; \
+		cd "../.."; \
+		'
 
 .PHONY: argocd
 argocd: argocd_gcp argocd_aws argocd_azure  ## Deploys ArgoCD
@@ -153,22 +163,35 @@ argocd_%:
 		done; \
 		'
 
-.PHONY: monitoring
-monitoring:  ## Deploys the TSB monitoring stack
-	@echo "Deploying TSB monitoring stack..."
-	@$(MAKE) k8s
+.PHONY: external-dns
+external-dns: external-dns_gcp  ## Deploys External DNS
+external-dns_%:
+	@echo "Deploying External DNS..."
+	@$(MAKE) $*_k8s
 	@/bin/sh -c '\
-		cd "addons/monitoring"; \
+		cd "addons/external-dns/$*/credentials"; \
 		terraform workspace select default; \
 		terraform init; \
-		terraform apply ${terraform_apply_args} -var-file="../../terraform.tfvars.json"; \
-		terraform output ${terraform_output_args} | jq . > ../../outputs/terraform_outputs/terraform-monitoring.json; \
+		terraform apply ${terraform_apply_args} -var-file="../../../../terraform.tfvars.json"; \
+		cd "../../../.."; \
+		index=0; \
+		jq -r '.$*_k8s_regions[]' terraform.tfvars.json | while read -r region; do \
+		echo "cloud=$* region=$$region cluster_id=$$index"; \
+		cd "addons/external-dns/$*"; \
+		terraform workspace new $*-$$index-$$region; \
+		terraform workspace select $*-$$index-$$region; \
+		terraform init; \
+		terraform apply ${terraform_apply_args} -var-file="../../../terraform.tfvars.json" -var=cluster_id=$$index; \
 		terraform workspace select default; \
-		cd "../.."; \
+		index=$$((index+1)); \
+		cd "../../.."; \
+		done; \
 		'
 
+##@ Clenaup
+
 .PHONY: destroy
-destroy: destroy_remote destroy_local
+destroy: destroy_remote destroy_local  ## Destroy the environment and the local Terraform state and cache
 
 .PHONY: destroy_remote
 destroy_remote:  ## Destroy the environment
@@ -220,3 +243,11 @@ destroy_tfcache:
 destroy_outputs:
 	rm -f outputs/*-kubeconfig.sh outputs/*-jumpbox.sh outputs/*-kubeconfig outputs/*.jwk outputs/*.pem
 	rm -f outputs/terraform_outputs/*.json
+
+##@ Others
+
+.PHONY: help
+help: Makefile  ## Prints this help
+	@awk 'BEGIN {FS = ":.*##"; printf "Usage:\n  make \033[36m<target>\033[0m\n"} \
+			/^[.a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } \
+			/^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) }' $(MAKEFILE_LIST)
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,7 @@ tsb_mp:  ## Deploys the TSB Management Plane
 		fqdn=`jq -r '.tsb_fqdn' ../../terraform.tfvars.json`; \
 		address=`jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" ../../outputs/terraform_outputs/terraform-tsb-mp.json`; \
 		terraform -chdir=../fqdn/$$cloud init; \
+		terraform -chdir=../fqdn/$$cloud apply ${terraform_apply_args} -var-file="../../../terraform.tfvars.json" -var=address=$$address -var=fqdn=$$fqdn; \
 		terraform workspace select default; \
 		cd "../.."; \
 		'
--- a/addons/external-dns/gcp/credentials/main.tf
+++ b/addons/external-dns/gcp/credentials/main.tf
@@ -1,0 +1,25 @@
+data "terraform_remote_state" "infra" {
+  backend = "local"
+  config = {
+    path = "../../../../infra/gcp/terraform.tfstate.d/gcp-${var.tsb_mp["cluster_id"]}-${var.gcp_k8s_regions[tonumber(var.tsb_mp["cluster_id"])]}/terraform.tfstate"
+  }
+}
+
+locals {
+  dns_project = endswith(var.tsb_fqdn, ".gcp.cx.tetrate.info") ? "dns-terraform-sandbox" : data.terraform_remote_state.infra.outputs.project_id
+}
+
+resource "google_service_account" "external_dns" {
+  project = local.dns_project
+  account_id = "${var.name_prefix}-external-dns"
+}
+
+resource "google_project_iam_member" "dns_admin" {
+  project = local.dns_project
+  role = "roles/dns.admin"
+  member = "serviceAccount:${google_service_account.external_dns.email}"
+}
+
+resource "google_service_account_key" "external_dns_key" {
+  service_account_id = google_service_account.external_dns.name
+}
--- a/addons/external-dns/gcp/credentials/outputs.tf
+++ b/addons/external-dns/gcp/credentials/outputs.tf
@@ -1,0 +1,8 @@
+output "dns_project" {
+  value = local.dns_project
+}
+
+output "dns_credentials" {
+  value     = base64decode(google_service_account_key.external_dns_key.private_key)
+  sensitive = true
+}
--- a/addons/external-dns/gcp/main.tf
+++ b/addons/external-dns/gcp/main.tf
@@ -20,7 +20,7 @@ module "external_dns" {
   k8s_client_token           = data.terraform_remote_state.infra.outputs.token
   dns_provider               = "google"
   tsb_fqdn                   = var.tsb_fqdn
-  dns_zone                   = replace(var.tsb_fqdn, "/^[^\\.]+\\./", "")  # Remove the first aprt of the fqdn
+  dns_zone                   = replace(var.tsb_fqdn, "/^[^\\.]+\\./", "")  # Remove the first part of the fqdn
   google_project             = data.terraform_remote_state.dns.outputs.dns_project
   google_service_account_key = data.terraform_remote_state.dns.outputs.dns_credentials
 }
--- a/addons/external-dns/gcp/main.tf
+++ b/addons/external-dns/gcp/main.tf
@@ -1,0 +1,26 @@
+data "terraform_remote_state" "infra" {
+  backend = "local"
+  config = {
+    path = "../../../infra/gcp/terraform.tfstate.d/gcp-${var.cluster_id}-${var.gcp_k8s_regions[var.cluster_id]}/terraform.tfstate"
+  }
+}
+
+data "terraform_remote_state" "dns" {
+  backend = "local"
+  config = {
+    path = "credentials/terraform.tfstate"
+  }
+}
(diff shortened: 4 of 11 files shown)


Makefile
```diff
@@ -153,22 +163,35 @@ argocd_%:
...
+	@echo "Deploying External DNS..."
+	@$(MAKE) $*_k8s
 	@/bin/sh -c '\
-		cd "addons/monitoring"; \
+		cd "addons/external-dns/$*/credentials"; \
 		terraform workspace select default; \
 		terraform init; \
-		terraform apply ${terraform_apply_args} -var-file="../../terraform.tfvars.json"; \
-		terraform output ${terraform_output_args} | jq . > ../../outputs/terraform_outputs/terraform-monitoring.json; \
+		terraform apply ${terraform_apply_args} -var-file="../../../../terraform.tfvars.json"; \
```
nacx, line 176: We do an initial run to create the credentials to manage DNS records, then iterate the clusters and install external-dns with those credentials.

gonzaloserrano (approved):

addons/external-dns/gcp/main.tf
```diff
@@ -0,0 +1,26 @@
...
+
+module "external_dns" {
+  source                     = "../../../modules/addons/external-dns"
+  cluster_name               = data.terraform_remote_state.infra.outputs.cluster_name
+  k8s_host                   = data.terraform_remote_state.infra.outputs.host
+  k8s_cluster_ca_certificate = data.terraform_remote_state.infra.outputs.cluster_ca_certificate
+  k8s_client_token           = data.terraform_remote_state.infra.outputs.token
+  dns_provider               = "google"
+  tsb_fqdn                   = var.tsb_fqdn
+  dns_zone                   = replace(var.tsb_fqdn, "/^[^\\.]+\\./", "")  # Remove the first part of the fqdn
```
smarunich, line 23: we must create a sub-zone per each cluster, i.e. the provided dns_zone should host a subzone thats with the cluster_name, i.e. $var.cluster_name-$var.dns_zone so it will keep the provided zones clean and manageable, especially if the public zones are being used

addons/external-dns/gcp/main.tf
```diff
@@ -0,0 +1,26 @@
...
+  backend = "local"
+  config = {
+    path = "../../../infra/gcp/terraform.tfstate.d/gcp-${var.cluster_id}-${var.gcp_k8s_regions[var.cluster_id]}/terraform.tfstate"
+  }
+}
+
+data "terraform_remote_state" "dns" {
+  backend = "local"
+  config = {
+    path = "credentials/terraform.tfstate"
```
smarunich, line 11: The credentials should not be part of `addons` - we should host all the dependencies to stand up the module under the given module and the actual "run" should only reference the module with the provided variables.

smarunich, line 11: we should limit the referencing of the states across the modules and do communicate the variables over.

addons/external-dns/gcp/credentials/main.tf
```diff
@@ -0,0 +1,25 @@
...
+  config = {
+    path = "../../../../infra/gcp/terraform.tfstate.d/gcp-${var.tsb_mp["cluster_id"]}-${var.gcp_k8s_regions[tonumber(var.tsb_mp["cluster_id"])]}/terraform.tfstate"
+  }
+}
+
+locals {
+  dns_project = endswith(var.tsb_fqdn, ".gcp.cx.tetrate.info") ? "dns-terraform-sandbox" : data.terraform_remote_state.infra.outputs.project_id
+}
+
+resource "google_service_account" "external_dns" {
```
smarunich, line 12: I think the right place for the "credentials" function it is actual "external_dns" function, however the external dns function should reference the pre-created subzones, i.e. each cluster external dns records should under the cluster name itself, to avoid the naming collisions between the clusters as well as to keep the configuration manageable.

smarunich, line 12: I think the right place for the "credentials" function it is actual "dns subzone that will provide not only zone reference but the credentials as well for external dns to use" function, however the external dns function should reference the pre-created subzones, i.e. each cluster external dns records should under the cluster name itself, to avoid the naming collisions between the clusters as well as to keep the configuration manageable.

smarunich (changes requested):
this is a perfect start however, we should decouple and make it manageable to reproduce across the three of the clouds, the first call to make:

1. `external-dns_dependecies` per cluster - we should create `subzone`  and related service account 
2. `external-dns` function that actually deploys the function 

(1) and (2) can be the same function or can be split, it is much easier to make a call when the implementation is being accounted/evaluated/written at least for two clouds. 

The logic should follow `provision dependencies` per given cluster > `export` the dependencies as variables, not just as a state > supply `variables` to the `external-dns` function to satisfy the install

Let me know if you want to re-write it so I won't sound like I am declining your changes... 

we should avoid functions like `credentials` as they basically a module inside of an environment, so to keep linear and manageable let's keep all the helpers inside the `modules`.

addons/external-dns/gcp/credentials/main.tf
smarunich: to keep ot simple lets have this helper function as part of the module that is designated to create a dns subzone and related service account per given cluster and output the variables as part of the defined module, it will be identical across the clouds too.

addons/external-dns/gcp/credentials/outputs.tf
```diff
@@ -0,0 +1,8 @@
+output "dns_project" {
+  value = local.dns_project
+}
+
+output "dns_credentials" {
```
smarunich, line 5: "credentials" function should not be just a credential function, but a subzone create function too with the given outputs, the outputs must be cloud agnostic.

modules/addons/external-dns/main.tf
```diff
@@ -0,0 +1,49 @@
...
+    name  = "txtOwnerId"
+    value = var.cluster_name
+  }
+
+  # Since we are enabling the Istio Gateway source, it is important to set this filter
+  # so we don't create DNS records for any random or internal hostnames defined in
+  # the TSB IngressGateways.
+  set {
+    name  = "domainFilters"
+    value = "{${var.dns_zone}}"
```
smarunich, line 37: the dns zone should be a cluster specific and created by the dependencies module that also includes creation of a service account.

modules/addons/external-dns/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+}
+
+# Provider specific variables. All of them should be optional to allow
+# configuring only the ones for the provider being used.
+
+variable "google_project" {
+  default = ""
+}
+
+variable "google_service_account_key" {
```
smarunich, line 29: depends how you want to twist, but we should account the implementation for the other two clouds as well, i.e. we cannot have a generic function that only can be used by GCP. However technically external-dns is a very unique function, so perhaps we have a cloud specific external-dns module that also takes care about managing the dependencies for the module itself

modules/addons/external-dns/main.tf
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  set {
+    name  = "sources"
+    value = "{service,ingress,istio-gateway}"
+  }
+
+  values = [templatefile("${path.module}/manifests/values-${var.dns_provider}.yaml.tmpl", {
```
smarunich, line 45: i understand why we are trying to do `dns_provider` here, however this is not necessary cloud agnostic approach here, so let's make a call if we want to have `external-dns` as a single module that accounts sub-zone creation as well service account management with the post-install function after it, instead of trying to do extra lookups that may drive unnecessary complexity, i.e. the code should be easily picked by the field team and customer... if needed

modules/addons/external-dns/main.tf
```diff
@@ -0,0 +1,49 @@
...
+  # so we don't create DNS records for any random or internal hostnames defined in
+  # the TSB IngressGateways.
+  set {
+    name  = "domainFilters"
+    value = "{${var.dns_zone}}"
+  }
+
+  set {
+    name  = "sources"
+    value = "{service,ingress,istio-gateway}"
```
smarunich, line 42: +istio-virtualservice ? technically both are supported to be mentioned together.

addons/external-dns/gcp/credentials/outputs.tf
smarunich: as more I think, we should just have a cloud specific `external-dns` function/module that include dependencies management and performs the install, and keeps credentials within the module (with no exposure)

Makefile
smarunich: yeah, let's do inside of the dependencies module as discussed earlier.

addons/external-dns/gcp/credentials/main.tf
nacx: I considered the subzone approach, but it has several cons:

There is already a predefined shared zone that is already preconfigured and records for that zone need to land there. It is common to expect that users may want to use the same domain for the MP and for the services they deploy for a consistent experience.
Enabling the subzone creation would require us to grant IAM admin roles (at least in gcp) to any user using this repo, and probably this is not what we want. We may, but I didn't want to impose that big permissions grant on a shared project now, to get the functionality going.
Note that is users are using a private project in gcp this is already creating records in the local DNS zone of their project.

I think reusing the existing zone makes sense. In GCP we can already accommodate other zones: we can use the shared one, or create a new public one in the user's project or even a private one. We just have to align the FQN module of the other providers to how we do stuff in GCP, instead of hacking this one.

addons/external-dns/gcp/credentials/outputs.tf
nacx: I already replied to the subzone part in the previous comment.

For the dependencies part, that's fine. I can move the credentials to `modules/gcp/dns-credentials` if needed.
I wanted to avoid doing a specific per-provider module because it is not needed and it would lead us to duplicate stuff: if we have a `module/gcp/external-dns` (like we have in the `register_fqdn`, we still have to have the per-provide thing in the add-on folder, because module paths cannot be parameterized, so we don't gain anything.

The current approach, instead, already has a completely generic module that requires little inputs and can be reused in every cloud.

addons/external-dns/gcp/main.tf
nacx: See the reasonning on the previous comment. This can be moved to the generic modules folder, but I think it's cleaner to keep one generic external-dns module for the reasoning above.

nacx: Yes, we should, but the credentials are only returned _on creation_. You can't get the private key with a data provider and you can't get it after wards.
Since we only configure DNS in one project, not once per cluster, we don't have any other option than storing the credential locally so that we can reference it for all clusters.

addons/external-dns/gcp/main.tf
nacx: Already replied to this previously.

modules/addons/external-dns/main.tf
nacx: Ditto.

modules/addons/external-dns/main.tf
nacx: Yes, both are supported but we only need the gateways, which is where the really exposed services are defined. We don't need VSs here.

modules/addons/external-dns/main.tf
nacx: There is no added complexity here. As I explained above, you can't parameterize module paths and you'll need a provider-specific module anyway in the add-ons folder, so there is no added complexity.
You have one generic module that requires the provider parameters and the credentials, and you call it from each provider module in the add-ons (that you have to create anyway) just with the provider-specific data. Less code, less modules tor season about.

modules/addons/external-dns/variables.tf
nacx: This is a generic module that should be expanded with the other cloud provider parameters.
I already replied to the module structure thing in previous comments.

Regarding supporting all clouds, yes, we want to support both, but it's not needed for this PR. [You already said this is an OK approach and that you'd do the rest](https://github.com/tetrateio/tetrate-service-bridge-sandbox/issues/184#issuecomment-1364102871) so I count on that after this PR is merged.

nacx:
>this is a perfect start however, we should decouple and make it manageable to reproduce across the three of the clouds, the first call to make:

Starting with one cloud, if everything is backwards-compatible is good enough, even more when adding new features no one is still using. And I rely on you already saying [this is an OK approach and that you'd take care of adding the other providers](https://github.com/tetrateio/tetrate-service-bridge-sandbox/issues/184#issuecomment-1364102871). I count on that after this PR is merged.

>external-dns_dependecies per cluster - we should create subzone and related service account
external-dns function that actually deploys the function
(1) and (2) can be the same function or can be split, it is much easier to make a call when the implementation is being accounted/evaluated/written at least for two clouds.

I think the subzone thing should be discussed. We have the option to use shared, public in the user's project or a private zone in GCP already, and this PR already accounts for that. Instead, I think the FQDN of the other providers should be aligned with this ability to support user zones, and this change keep in alignment with the FQDN module (which is how it works right now).

>The logic should follow provision dependencies per given cluster > export the dependencies as variables, not just as a state > supply variables to the external-dns function to satisfy the install

That's great, but can't be done as the private key for the credentials can't be retrieved in a data provider; it's only returned at create time. Since the resource is created only once, but needs to be accessed by all clusters, it needs to be stored locally.

>Let me know if you want to re-write it so I won't sound like I am declining your changes...

Definitely no. We already discussed how collaboration works and we should just stick to our [protocol guidelines](https://github.com/tetrateio/protocol/blob/main/engineering/code-reviews.md).

>we should avoid functions like credentials as they basically a module inside of an environment, so to keep linear and manageable let's keep all the helpers inside the modules.

I can move the credentials module to the generic module folder. However, this is a dependency we have to manage a bit differently given the nature of all the differences between cloud providers. This is exactly the same as the FQDN module. If you look, we have the `tsb/fqdn` module that we manage separately, for a reason. If this is an OK approach for the TSB module, it should be an OK approach for an add-on.

addons/external-dns/gcp/credentials/main.tf
smarunich: let's not get centric to GCP (i.e. I will not list the reasons on a public repo ), the cons are you listing do not account very basic collisions (i.e. at least two clusters) across multiple deployments leveraging the common dns zone, on top GCP private zone is a corner case deployment as of now (and not extended across the multiple clouds, i.e. all the demos we do - do include multiple clouds not GCP is the least preferred option), also to keep in mind the demos we do - do use public zones so after can be easily reached and entertained by the customers as well. I agree on allowing to push external-dns to the common parent zone too, but it should not be a default option and we should implement more safe model to start, I can re-write the PR if you want. 

IAM is a necessary evil to fight which is very hard to avoid (and the permissions for DNS management for example under GCP project already includes it, for AWS and Azure)  as OpenShift deployment does require a dedicated sub-zone. 

We need to do a sub-zone to overcome:

1. naming collisions, i.e. "vmgateway" service under dns-zone1/demo-ignasi1 will be affected by a new deployment under the same zone by another deployment dns-zone1/demo-sergey1
2. keeping the underlay DNS infrastructure clean
3. follow external-dns common deployment patterns, i.e. sub-zone per the given cluster which also would be an easy to describe to the end consumer, and account multiple clusters

addons/external-dns/gcp/credentials/outputs.tf
smarunich: it is a hard point to argue, as you are implementing it only for GCP stage so it is hard to technically prove it is generic, the easiest way to get to the common, it is to compare and merge the three options when drafted, so if you don't mind - can you please store all the steps under the given `module/gcp/external-dns` folder - technically at least it will account to the general stands of how I am trying to keep this repo.

addons/external-dns/gcp/main.tf
smarunich: `module/gcp/external-dns ` comment

smarunich: so anyway this is a function of "output" for the given module, account we follow `module/gcp/external-dns ` comment

addons/external-dns/gcp/main.tf
smarunich: accounting we follow `module/gcp/external-dns ` with `sub-zone` approach.

modules/addons/external-dns/main.tf
smarunich: accounting we follow module/gcp/external-dns  with sub-zone approach.

modules/addons/external-dns/main.tf
smarunich: out of curiosity, is there a harm of `istio-virtualservice` be there? i.e. we decided to enforce an exact names under gateway, however if `TID` or `PLZ` will be used - i usually do see `wildcards` on gateway in OSS :) world. so keeping it there would it harm?

modules/addons/external-dns/main.tf
smarunich: if the dependencies are managed within the module - there is no need even to pass them externally so having "credentials" in outbound way like it is defined now, does create a one more level of execution to account, even though we manage it via `Makefile` but this is not an ideal thing that we want to do, and why it is bad to have all the dependencies under the module that only needs it?  i.e. accounting we follow module/gcp/external-dns  with sub-zone approach.

modules/addons/external-dns/variables.tf
smarunich: i think it is important to keep the intention of this repo in line to the point behind it is there to support multi-cloud demos, when integration is being done I also do think how can I fit AWS or Azure in the same time, not later as technically we always do multi-cloud demos so I cannot ignore the matter, so I think it will be fair for the important function like this ... consider how it can be done to account at least Azure or AWS, so the PR does need to include AWS or Azure, but it should aligned for multi-cloud usage pattern.

addons/external-dns/gcp/credentials/main.tf
nacx: 1. I disagree with the collision's argument. If a user has N clusters is very understandable they won't be assigning the same DNS name to each VMgateway, so that shouldn't be an issue. Note that the VM gateway gets the DNS name from the config you set in the CP resource, and we can expect users to not put the same name there in all clusters (as that's a publicly resolvable name), so the argument of the name collision does not apply.
2. If the AWS and Azure providers are as flexible as GCP and support the notion of Private, Public and Shared zone, the DNS infra remains clean. Note that you're only talking about Private, and you're assimilating Public to Shared in your argument, but you're missing one. The Publc zone individually managed is a Public zone different than the Shared one. This already meets the use case you're requesting of keeping the DNS infra clean and isolated. It's just that AWS and Azure are not yet as flexible as Google; let's fix that instead of doing stuff that only applies to an add-on.
3. Given the above two points and that users already would have a way to choose between Private, Public (owned only by them) and Shared, this looks more like an improvement that could be perfectly implemented in another pull request but not a major blocker for this one.

addons/external-dns/gcp/credentials/outputs.tf
nacx: >it is a hard point to argue

But still ¯\\_(ツ)_/¯ 😂

>how I am trying to keep this repo.

this repo is no longer about _you_.

modules/addons/external-dns/main.tf
nacx: You keep assuming that this can be all done in one module, and we can't.

Credentials must be created _once_, external-dns must be installed N times reusing those credentials created once, so these two cases are better modeled in two different modules. No matter where you place the code, in the add-ons folder or in the generic modules you still need two modules.

modules/addons/external-dns/variables.tf
nacx: This is not core funcitonality; it is an add-on and it's not mandatory for demos. You have survived demos till now without this, so, why shouldn't we be able to implement it incrementally?

Also, the code in this PR is fully prepared to be replicated in other clouds without major changes. it doesn't follow the structure you had envisioned, but there are good reasons and arguments to have it this way. The only thing that could change is moving the `credentials` module inside the `modules` folder, but it still needs to be a separate module with a different execution, because, as I said before, it has to be executed only once, but external-dns needs to be installed for each cluster.

addons/external-dns/gcp/credentials/main.tf
nacx: >and the permissions for DNS management for example under GCP project already includes it, for AWS and Azure

This would mean coupling the extenral-dns add-on to some internals of how the MP is deployed and the fact that it configures DNS. Coupling this is a very bad design. If at some point we change the internals of how we deploy the MP, we would break extenral-dns. breaking add-ons when some internals of the MP module change is not a good design, especially if they can be completely decoupled, which is how it is right now.

modules/addons/external-dns/main.tf
smarunich: there is `outputs` folder to store generated artefacts and module function can store and lookup file over there.

addons/external-dns/gcp/credentials/main.tf
smarunich: so i am bias to the approach, as this is a field facing repo, so somebody has to take a lead on the decisions, and this is my call to tell you that you may disagree, but this is the reasons as we see them as a field team, so if you want to have GCP fork - we can always have a branch that describes and fits your needs, but main story around this repo is not you, but field.

addons/external-dns/gcp/credentials/outputs.tf
smarunich: i love the comments and energy, but @nacx this is not a monorepo, so you cannot lead by authority, so please do take into consideration the fact this is not a toy for you to overtake, but a field facing repo that has a consumers beside you and was built as my personal toy project which appeared to be of use to the people so I will still do :) pet this repo, until I am here so :) consensus and balance that what I need you to account here.

modules/addons/external-dns/variables.tf
smarunich: the execution can store the output in `outputs` and execute only once based on the existence of the file, after on file can retrieved and stored a `data` for the given `module`.

smarunich (changes requested):
having a good time here

smarunich:
apparently the reviews are not showing up in the main thread, so have to be followed under the files review.

nacx:
To summarize:

>so i am bias to the approach, as this is a field facing repo, so somebody has to take a lead on the decisions, and this is my call to tell you that you may disagree, but this is the reasons as we see them as a field team, so if you want to have GCP fork - we can always have a branch that describes and fits your needs, but main story around this repo is not you, but field.

>i love the comments and energy, but @nacx this is not a monorepo, so you cannot lead by authority, so please do take into consideration the fact this is not a toy for you to overtake, but a field facing repo that has a consumers beside you and was built as my personal toy project which appeared to be of use to the people so I will still do :) pet this repo, until I am here so :) consensus and balance that what I need you to account here.

Even though this works perfectly fine and is a good starting point (for an add-on, not core functionality), instead of favoring collaboration and an incremental approach, this is going to be pushed back unless it aligns with your mental model.
Looks like this repo, given its history has a benevolent dictator, and there is no intention to change that, or at least to have meaningful arguments beyond "this is not yours" or "believe me". benevolent dictators may be great, but that does not favor collaboration, and gatekeeping on this repo towards opinions, instead of favouring incremental progress seem to be what's is being imposed. ACK. but I encourage you to revisit the decision of moving this and getting it back to your personal account if this is going to be the case.

I'm closing this as +30 comments for an add-on is already too much of a discussion that is not worth the effort.

PR #218, opened 2023-01-20
Cleanup unused providers

nacx:
* Cleanup unused providers.
* Removes the `keycloak` module from the addons folder as it is broken. It still contained data references to a structure that no longer exists and is broken. I've not deleted the generic keycloak modules, but the end-user facing add-on only created confusion.
* Cleans up a bit the eshop gitops config, leveraging argocd waves to have the files nicely split.
* Adds the `argocd_include_example_apps` variable to allow installing ArgoCD without the example applications. This is very useful to have ArgoCD installed in all the clusters but without applications, so users can just configure their own apps for their own repos (by default example apps will still be installed).


smarunich (approved):

smarunich (approved):

smarunich:
agree, technically we should have keycloak as `argocd` construct,

PR #226, opened 2023-02-08
updates to account new aws tagging requirements

smarunich:
updates to account new aws tagging requirements
added random_id to name_prefix for each aws object
addressed https://github.com/tetrateio/tetrate/issues/15368
updates to account GCP and Azure tagging requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Please refer to [tfvars collection](/tfvars_collection) for more examples, i.e. 
 {
   "name_prefix": "<YOUR UNIQUE PREFIX NAME TO BE CREATED>",
   "tsb_fqdn": "<YOUR UNIQUE PREFIX NAME TO BE CREATED>.cx.tetrate.info",
-  "tsb_version": "1.5.0",
+  "tsb_version": "1.6.0",
   "tsb_image_sync_username": "<TSB_REPO_USERNAME>",
   "tsb_image_sync_apikey": "<TSB_REPO_APIKEY>",
   "tsb_password": "Tetrate123",
@@ -62,7 +62,9 @@ Please refer to [tfvars collection](/tfvars_collection) for more examples, i.e. 
   "tsb_org": "tetrate",
   "aws_k8s_regions": [],
   "azure_k8s_regions": ["eastus"],
-  "gcp_k8s_regions": ["us-west1", "us-east1"]
+  "gcp_k8s_regions": ["us-west1", "us-east1"],
+  "tetrate_owner": "username",
+  "tetrate_team": "function:team"
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -31,10 +31,17 @@ The `Makefile` in this directory provides ability to fastforward to anypoint of 
 ## Prerequisites
 
 - terraform >= 1.3.6
-- AWS role configured and assumed (Route53 is used for TSB MP FQDN)
+- (optional) AWS role configured and assumed (Route53 is used for TSB MP FQDN)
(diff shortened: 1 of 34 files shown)


smarunich:
will update azure and gcp to follow

README.md
```diff
@@ -62,7 +62,9 @@ Please refer to [tfvars collection](/tfvars_collection) for more examples, i.e.
   "tsb_org": "tetrate",
   "aws_k8s_regions": [],
   "azure_k8s_regions": ["eastus"],
-  "gcp_k8s_regions": ["us-west1", "us-east1"]
+  "gcp_k8s_regions": ["us-west1", "us-east1"],
+  "tetrate_owner": "username",
+  "tetrate_team": "function:team"
```
nacx: [nit] Add a link to the document that lists the allowed values, as the tag keys don't accept random strings.

modules/aws/base/main.tf
```diff
@@ -1,18 +1,10 @@
...
-  lower   = true
-  upper   = false
-  numeric = false
-}
 resource "aws_vpc" "tsb" {
   cidr_block           = var.cidr
   enable_dns_hostnames = true
-  tags = {
-    Name            = "${var.name_prefix}-${random_string.random_prefix.result}_vpc"
+  tags = merge({
```
nacx: [nit] It's probably safer to configure the default tags in the provider declaration (like I do in [my demo VM-Onboarding setup](https://github.com/tetrateio/eshop-demo/blob/main/vm/aws/provider.tf#L15-L17)) instead of explicitly declaring and merging them. This will reduce the chances of leaving resources untagged by mistake.

smarunich: thank you, unfortunately volume and launch template don't respect default tags, neither GCP or Azure supports the default tags/labels, I have updated AWS to follow the way you suggested.

README.md
smarunich: added to prereqs under README.md

smarunich:
updated tagging for Azure and GCP as well

shamusx (approved):

modules/aws/base/main.tf
nacx: Yup, that's unfortunate. I also faced the volume_tags thing!

README.md
nacx: [nit] Since it is important that we enforce the right values here, would it be better to add a value like:

```
"tetrate_owner": "Change me! (URL)",
```

To avoid an easy copy/paste mistake? In AWS policies can't validate the _value_ of the tags, so we should try to minimize the chances invalid values are set just because of copy/paste.

modules/addons/ratelimit/main.tf
```diff
@@ -9,7 +9,7 @@ provider "helm" {
 resource "random_password" "redis" {
   length = 16
   # Do not use the ':' character here as it will be read as Basic Auth and tokenized as <user>:<password>
-  override_special = "!@#$%&*()-_=+[]{}<>?"
+  special = false
```
nacx, line 12: Are there other characters apart from the `:` that cause trouble?

smarunich: If I recall correct, as I was multi-tasking with the customer - I had password with `{`, so decided to keep it simple, also searched around looks it is a common issue, found a reference here as well: https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/auth.html - decided to keep it simple, as technically 16 symbols should cover the password complexity requirements.

nacx (approved):

PR #231, opened 2023-02-14
Migrate from `*.cx.tetrate.info` to `*.sandbox.tetrate.io`

smarunich:
Migrate from `*.cx.tetrate.info` to `*.sandbox.tetrate.io`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,8 @@ tsb_mp:  ## Deploys MP
 	@echo "Deploying TSB Management Plane..."
 	@/bin/sh -c '\
 		cloud=`jq -r '.tsb_mp.cloud' terraform.tfvars.json`; \
+		# to account sandbox.tetrate.io as a default domain for GCP
+		dns_provider=`jq -r '.tsb_fqdn' terraform.tfvars.json | cut -d"." -f2 | sed 's/sandbox/gcp/g'`;\
 		cd "tsb/mp"; \
 		terraform workspace select default; \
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -93,6 +93,8 @@ tsb_mp:  ## Deploys MP
 	@echo "Deploying TSB Management Plane..."
 	@/bin/sh -c '\
 		cloud=`jq -r '.tsb_mp.cloud' terraform.tfvars.json`; \
+		# to account sandbox.tetrate.io as a default domain for GCP
+		dns_provider=`jq -r '.tsb_fqdn' terraform.tfvars.json | cut -d"." -f2 | sed 's/sandbox/gcp/g'`;\
```
nacx, line 97: Why not keep using the `cloud` variable? We configure DNS in the `tsb_mp` target, so probably just reading the cloud variable is less error-prone, if we have to change DNS naming?

nacx: BTW, Why are we special casing gcp here?

smarunich: 1. `tsb_mp.cloud` can be AWS, but for example `tsb_fqdn` can be managed by GCP, i.e. `tsb_fqdn` can be equal to `test.gcp.sandbox.io`
2. `sandbox.tetrate.io` is hosted by GCP, so in case if we want to hide the cloud name for the demo purposes in the TSB MP fqdn

nacx (approved):

Issue #241, opened 2023-02-23
When using development versions jumpbox user-data is not idempotent


nacx:
When using development versions, the following is used to  get the credentials to the development image repository: https://github.com/tetrateio/tetrate-service-bridge-sandbox/blob/d0fb41a6ebb6f7c5f425fb19bd838fd1f861395e/modules/internal_registry/internal-cr-token.sh#L9

This is then injected to the jumpbox `user_data`, but since the tokens are short-lived and different every time, this causes an unnecessary reconcile of the jumpbox. These reconciliations can cause the jumpbox to get a different public IP address, breaking the existing ssh scripts.

PR #242, opened 2023-02-23
Fix AWS cleanup and jumbox idempotence

nacx:
Adds several fixes to properly cleanup resources in AWS, and changes a bit its user_data to make sure it doesn't change on every run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,6 +179,7 @@ destroy_remote:  ## Destroy the environment
 		address=`jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" outputs/terraform_outputs/terraform-tsb-mp.json`; \
 		cd "tsb/fqdn/$$cloud"; \
 		terraform destroy ${terraform_apply_args} -var-file="../../../terraform.tfvars.json" -var=address=$$address -var=fqdn=$$fqdn; \
+		[ $$? -ne 0 ] && exit 1; \
 		rm -rf terraform.tfstate.d/; \
 		rm -rf terraform.tfstate; \
 		cd "../../.."; \
@@ -200,8 +201,7 @@ destroy_%:
 		cd "infra/$*"; \
 		terraform workspace select $*-$$index-$$region; \
 		terraform destroy ${terraform_destroy_args} -var-file="../../terraform.tfvars.json" -var=$*_k8s_region=$$region -var=cluster_id=$$index; \
-		terraform workspace select default; \
-		terraform workspace delete ${terraform_workspace_args} $*-$$index-$$region; \
+		[ $$? -eq 0 ] && terraform workspace select default && terraform workspace delete ${terraform_workspace_args} $*-$$index-$$region; \
 		index=$$((index+1)); \
 		cd "../.."; \
 		done; \
@@ -219,5 +219,5 @@ destroy_tfcache:
 
 .PHONY: destroy_outputs
 destroy_outputs:
-	rm -f outputs/*-kubeconfig.sh outputs/*-jumpbox.sh outputs/*-kubeconfig outputs/*.jwk outputs/*.pem
+	rm -f outputs/*-kubeconfig.sh outputs/*-jumpbox.sh outputs/*-kubeconfig outputs/*.jwk outputs/*.pem outputs/*-cleanup.sh
 	rm -f outputs/terraform_outputs/*.json
--- a/modules/aws/jumpbox/aws_cleanup.sh.tmpl
+++ b/modules/aws/jumpbox/aws_cleanup.sh.tmpl
@@ -2,6 +2,12 @@
 
 export REGION="${region}"
 export VPC_ID="${vpc_id}"
+export REPO_PREFIX="${registry_name}"
+
+echo 'Destroying ECR repositories...'
+for repo in $(aws ecr describe-repositories --region $REGION --query 'repositories[*].repositoryName' --no-paginate --output text); do
+    [[ "$repo" =~ ^$REPO_PREFIX/.* ]] && aws ecr delete-repository --repository-name $repo --region $REGION --force
+done
 
 echo 'Destroying K8s ELBs...'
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -179,6 +179,7 @@ destroy_remote:  ## Destroy the environment
 		address=`jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" outputs/terraform_outputs/terraform-tsb-mp.json`; \
 		cd "tsb/fqdn/$$cloud"; \
 		terraform destroy ${terraform_apply_args} -var-file="../../../terraform.tfvars.json" -var=address=$$address -var=fqdn=$$fqdn; \
+		[ $$? -ne 0 ] && exit 1; \
```
nacx, line 182: Do not delete the directories if the destroy operation failed. Since the entire block runs as a single command, `make` doesn't abort and continues anyway, so we need to explicitly short-circuit that to avoid losing the status when stuff is still there in the cloud provider.

This allows retrying the destroy operation if it failed, for example, because of an expired token.

Makefile
```diff
@@ -200,8 +201,7 @@ destroy_%:
 		cd "infra/$*"; \
 		terraform workspace select $*-$$index-$$region; \
 		terraform destroy ${terraform_destroy_args} -var-file="../../terraform.tfvars.json" -var=$*_k8s_region=$$region -var=cluster_id=$$index; \
-		terraform workspace select default; \
-		terraform workspace delete ${terraform_workspace_args} $*-$$index-$$region; \
+		[ $$? -eq 0 ] && terraform workspace select default && terraform workspace delete ${terraform_workspace_args} $*-$$index-$$region; \
```
nacx, line 204: Do not delete the workspace if the destroy operation failed. Since the entire block runs as a single command, `make` doesn't abort and continues anyway, so we need to explicitly short-circuit that to avoid losing the status when stuff is still there in the cloud provider.

This allows retrying the destroy operation if it failed, for example, because of an expired token.

modules/aws/jumpbox/aws_cleanup.sh.tmpl
```diff
@@ -2,6 +2,12 @@
 
 export REGION="${region}"
 export VPC_ID="${vpc_id}"
+export REPO_PREFIX="${registry_name}"
+
+echo 'Destroying ECR repositories...'
+for repo in $(aws ecr describe-repositories --region $REGION --query 'repositories[*].repositoryName' --no-paginate --output text); do
+    [[ "$repo" =~ ^$REPO_PREFIX/.* ]] && aws ecr delete-repository --repository-name $repo --region $REGION --force
+done
```
nacx, line 10: in AWS, the `tctl sync` will create one repository for each image, but that's not managed by terraform and was never cleaned up.
This code makes sure all repositories that could have been created are properly removed.

modules/aws/jumpbox/jumpbox.userdata
```diff
@@ -29,7 +29,7 @@ write_files:
       #!/bin/bash
       curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
       install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
-      ${docker_login}
+      aws ecr get-login-password --region ${region} | docker login --username AWS --password-stdin ${registry}
```
nacx: This is moved from the `.tf` file to make user_data idempotent.

modules/aws/jumpbox/jumpbox.userdata
```diff
@@ -39,7 +39,9 @@ write_files:
         helm pull tetrate-tsb-helm/managementplane --version ${tsb_version} --devel
         tar zxvf managementplane-${tsb_version}.tgz
         for IMAGE in $(cut -d/ -f2- managementplane/images.txt)
-        do 
+        do
+          IMGNAME=$(echo $IMAGE | cut -d: -f1)
+          aws ecr create-repository --repository-name ${registry_name}/$IMGNAME --region ${region} || true
```
nacx, line 44: Repositories need to exist before pushing. this is something `tctl image sync` does internally, but when manually copying needs to be done explicitly. The `|| true` is to avoid having the command fail if the repositories exist (on eventual Jumbox recreations).

Makefile
gonzaloserrano: maybe we could pass instead the `-e` flag to the `sh` invocation to stop the execution on first error? with `@/bin/sh -e -c`

gonzaloserrano (approved):

Makefile
nacx: This will attempt to destroy stuff from cloud providers in a loop, and we want it to be able to destroy as much as possible, so I'd rather not stop the loop but let it continue without messing the state for stuff that couldn't be deleted.

modules/aws/jumpbox/jumpbox.userdata
smarunich: @nacx i feel like the https://github.com/tetrateio/tetrate-service-bridge-sandbox/blob/33e90ef57aa1135fd78fd97696739f98f223fb99/modules/aws/jumpbox/jumpbox.userdata#L45 is the issue, as it does include internal token that is being dynamically generated each run, but docker_login is static in the nature when file is rendered...

I am checking it now...

smarunich: I just checked indeed, it is the root cause, so we don't need to strip off docker_login. Technically keeping it there, minimizes the drift between the environments, do you mind rolling back to retain docker_login?

As we discussed on slack, perhaps the easiest way it is to have `aws_eip` elastic ip assigned to jumpbox to retain the SSH target.

Makefile
smarunich: thank you

modules/aws/jumpbox/jumpbox.userdata
nacx: Ah, right, actually I had this change before, but then got to the same conclusion and reported https://github.com/tetrateio/tetrate-service-bridge-sandbox/issues/241. I'll revert these lines back.

nacx: Done

PR #246, opened 2023-02-24
Cache the internal CR token to avoid jumbox reconciliations

### LINKED ISSUE #241: When using development versions jumpbox user-data is not idempotent

nacx:
Fixes https://github.com/tetrateio/tetrate-service-bridge-sandbox/issues/241

@smarunich this is an absolute hack, but WDYT about this?

The idea is that just because that module uses an external script and always runs, it's altering the suer-data and causing the jumpbox to reconcile, when that's useless because cloud-init is run only on the first boot; changes to user-data wouldn't even be considered.

This change basically caches the initial token in the local filesystem to return it in subsequent runs, and avoid that annoying reconcile that breaks the development version usage.

I also considered adding it as an output variable of the module and propagate the output so it can be cached in the terraform state, and read with the `terraform_remote_state` data provider, but that would fail on the first run when the key still does not exist, so I opted for the local (.git-ignored) file hack. WDYT?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,8 +210,7 @@ destroy_%:
 
 .PHONY: destroy_tfstate
 destroy_tfstate:
-	find . -name terraform.tfstate.d -exec rm -rf {} +
-	find . -name terraform.tfstate -delete
+	find . -name *tfstate* -exec rm -rf {} +
 
 .PHONY: destroy_tfcache
 destroy_tfcache:
--- a/modules/aws/jumpbox/main.tf
+++ b/modules/aws/jumpbox/main.tf
@@ -230,6 +230,14 @@ data "aws_availability_zones" "available" {}
 module "internal_registry" {
   source      = "../../internal_registry"
   tsb_version = var.tsb_version
+  # The internal registry token is needed only if the TSB version is a development version, and only once when the
+  # jumpbox bootstraps the first time. It is not needed later as all images are already pushed to the registry (and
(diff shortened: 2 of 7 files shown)


smarunich:
I think it is good, technically we can expand to use https://developer.hashicorp.com/terraform/language/state/remote-state-data#defaults just for the sake... interesting if it works  using `default` value :) ... or if that does not work simply the default empty using `try` https://developer.hashicorp.com/terraform/language/functions/try ... but also an experiment...  to leverage terraform_remote_state, but not a big deal in my opinion for this case.

nacx:
I've tried the remote_state thing but it doesn't work. even if I configure the defaults or use the try block, it fails on the first run with:
```
No stored state was found for the given workspace in the given backend.
```
I don't think this can be achieved with the state as I haven't found a way to have the data provider not fail if the file does not exist.

smarunich (approved):

PR #247, opened 2023-02-27
support for `external-dns` as `addon` for all three cloud providers

### LINKED ISSUE #222: Add an add-on to install and configure external-dns in all clusters
Add an add-on to install and configure external-dns in all clusters.
The add-on should allow users to configure DNS names for their applications and TSB infrastructure (vmgateway) in custom subzones or the existing shared zone.
As an optional add-on, it must not be coupled to the internals of the Management Plane deployment.

smarunich:
1. The sub-domain `$cluster_name.$external_dns_$cloud_dns_zone` for example `*.aks-r161rc0p3-eastus-0.azure.sandbox.tetrate.io`
2. The credentials for service account are done under the `external-dns` module itself
3. Support for all three cloud providers
4. external-dns is implemented as addon


infra/gcp/variables.tf
```diff
@@ -129,4 +129,24 @@ locals {
...
+variable "external_dns_annotation_filter" {
+  default = ""
+}
+
+variable "external_dns_label_filter" {
+  default = ""
+}
+
+variable "external_dns_sources" {
+  default = "service,ingress,istio-gateway"
```
nacx: I'd leave only `service` until we address the issue to help customize this. This will get in the way of T1 and T2, and for our demo environments it is easy enough for us to just label the `install.Ingress` resources. WDYT?

infra/gcp/main.tf
```diff
@@ -68,3 +68,18 @@ module "gcp_k8s" {
   tags               = local.default_tags
   depends_on         = [module.gcp_jumpbox[0]]
 }
+
+module "external_dns" {
+  source                     = "../../modules/gcp/external-dns"
+  name_prefix                = "${var.name_prefix}-${var.cluster_id}"
+  cluster_name               = var.cluster_name == null ? "gke-${var.gcp_k8s_region}-${var.name_prefix}" : var.cluster_name
```
nacx: [nit] let's unify code and use coalesce here as well?

infra/gcp/main.tf
```diff
@@ -68,3 +68,18 @@ module "gcp_k8s" {
...
+  k8s_host                   = module.gcp_k8s[0].host
+  k8s_cluster_ca_certificate = module.gcp_k8s[0].cluster_ca_certificate
+  k8s_client_token           = module.gcp_k8s[0].token
+  project_id                 = coalesce(var.gcp_project_id, google_project.tsb[0].project_id)
+  dns_zone                   = var.external_dns_gcp_dns_zone
+  sources                    = var.external_dns_sources
+  annotation_filter          = var.external_dns_annotation_filter
+  label_filter               = var.external_dns_label_filter
+  external_dns_enabled       = var.external_dns_enabled == true ?  true : false
```
nacx:
```suggestion
  external_dns_enabled       = var.external_dns_enabled
```

modules/gcp/external-dns/main.tf
```diff
@@ -0,0 +1,103 @@
...
+  project      = var.project_id
+  account_id   = "${var.name_prefix}-external-dns"
+}
+
+resource "google_dns_managed_zone" "cluster" {
+  count      = var.external_dns_enabled == true ? 1 : 0
+  project    = var.project_id
+  name       = "${var.cluster_name}-${local.zone_name}"
+  dns_name   = "${var.cluster_name}.${local.dns_name}."
+}
```
nacx: How will this work? Are users supposed to create their DNS names with the cluster name prefix?

infra/gcp/main.tf
smarunich: done, updated few more of iterations.

infra/gcp/main.tf
smarunich: lol, indeed sunday night sideeffects :), fixed.

nacx:
I'm not sure about the overall structure. External DNS is clearly an add-on. Actually, it is an even more clear add-on than Elasticsearch or the rate limiting server, and the structure should clearly reflect that and that external-dns should be moved to the addons folder.
I mean, this structure seems to favour convenience versus correctness, and that will lead to make the overall repo harder to reason about and potentially evolve, as we start coupling things.

infra/gcp/variables.tf
smarunich: yeah, fair enough.

modules/gcp/external-dns/main.tf
smarunich: at this stage, I have tested public zones as `gcp.sandbox.tetrate.io` and `azure.sandbox.tetrate.io`, i am doing `aws.sandbox.tetrate.io`, i have put the placeholders to support private zones, I think we should sort out private zones  management before merging this PR.

smarunich:
> I'm not sure about the overall structure. External DNS is clearly an add-on. Actually, it is an even more clear add-on than Elasticsearch or the rate limiting server, and the structure should clearly reflect that and that external-dns should be moved to the addons folder. I mean, this structure seems to favour convenience versus correctness, and that will lead to make the overall repo harder to reason about and potentially evolve, as we start coupling things.

agree, my bad,  I have moved the `external-dns` modules  under `addons`.

smarunich:
added support for AWS
added interval for external-dns refresh as a variable, defaults to `5s` 
added tags

infra/aws/main.tf
```diff
@@ -52,8 +52,27 @@ module "aws_k8s" {
...
   vpc_subnets  = module.aws_base[0].vpc_subnets
   name_prefix  = "${var.name_prefix}-${var.cluster_id}-${random_string.random_id.result}"
-  cluster_name = var.cluster_name == null ? "eks-${var.aws_k8s_region}-${var.name_prefix}" : var.cluster_name
+  cluster_name = coalesce(var.cluster_name, "eks-${var.aws_k8s_region}-${var.name_prefix}")
   output_path  = var.output_path
   tags         = local.default_tags
   depends_on   = [module.aws_jumpbox[0]]
 }
+
+module "external_dns" {
```
nacx: This still doesn't feel right. I mean, if it is an _optional addon_, it shouldn't be part fo the main infra module. I expect consumers to be able to do a `make external-dns` the same way they do a `make monitoring` or `make argocd`.
By coupling this here we are just dealing with optional add-ons in two different ways. On the one hand, we have add-ons that appear in the main infra module and you control them with a flag, on the other hand, we have proper add-ons that you explicitly install. This just causes confusion to contributors like me as there is no clear way of adding optional components and makes it hard to evolve the entire thing as now the main modules are coupled to ext-dns no matter what.

I really think this shouldn't be here and external-dns be installed with its dedicate make target like argo or the monitoring stack. This looks more that we're trading developer convenience (at the time of writing the PR) for correctness of the repo structure and that's only going to make things harder to maintain and evolve in the future.

smarunich: I hear you, as I have been also considering below, so in a way looked to DNS as a required infra...

The only reason it is still there, it is desire to get rid of "register_fqdn" step and perhaps have a common approach for dns records management... and eventually have CNAME for UI access... i.e. make it as baseline component, like we have cert-manager.

We need DNS for:

1. TSB MP endpoint (includes XCP)
2. VM onboarding
3. GWs

Even though we can have a separate pointer to external-dns install, I  will have it done.

nacx: That makes sense

modules/addons/aws/external-dns/main.tf
```diff
@@ -0,0 +1,66 @@
...
+}
+
+resource "aws_route53_record" "ns" {
+  count   = local.shared_zone && var.external_dns_enabled == true ? 1 : 0
+  zone_id = data.aws_route53_zone.shared[0].zone_id
+  name    = aws_route53_zone.cluster[0].name
+  type    = "NS"
+  ttl     = 300
+  records = aws_route53_zone.cluster[0].name_servers
+}
```
nacx, line 18: For my understanding, why do you need this? Isn't it enough to have the custom DNS zone?

nacx: Discussed offline. There are limitations in the DNS names we can assign with this model, as users will be required to include the cluster names in the DNS names, but I think it's ok to merge this and iterate on it once merged.

nacx (dismissed):

infra/aws/main.tf
```diff
@@ -52,8 +52,27 @@ module "aws_k8s" {
...
+  region                     = var.aws_k8s_region
+  vpc_id                     = module.aws_base[0].vpc_id
+  dns_zone                   = var.external_dns_aws_dns_zone
+  sources                    = var.external_dns_sources
+  annotation_filter          = var.external_dns_annotation_filter
+  label_filter               = var.external_dns_label_filter
+  interval                   = var.external_dns_interval
+  tags                       = local.default_tags
+  oidc_provider_arn          = module.aws_k8s[0].oidc_provider_arn
+  external_dns_enabled       = var.external_dns_enabled
```
nacx: Here and in the others, what about not propagating the `enabled` flag inside, and just use it as a `count` in this root module declaration? You control here at once if the module is applied or not, instead of having to "count" on every single resource inside the extdns module.

smarunich: as we do have helm provider configuration under the module... https://github.com/tetrateio/tetrate-service-bridge-sandbox/blob/6cc9d2410b3f7f7944665733fa9d5768aa996860/modules/addons/azure/external-dns/main.tf#L47 

in a way it goes to the point of externalizing... provider configuration away from module...

```
╷
│ Error: Module is incompatible with count, for_each, and depends_on
│
│   on main.tf line 55, in module "external_dns":
│   55:   count                      =  var.external_dns_enabled == true ? 1 : 0
│
│ The module at module.external_dns is a legacy module which contains its own local provider configurations, and so calls to it may not use the count, for_each, or depends_on arguments.
│
│ If you also control the module "../../modules/addons/azure/external-dns", consider updating this module to instead expect provider configurations to be passed by its caller.
╵
╷
```

nacx: ok, not a big deal then, let's keep it as-is

Makefile
```diff
@@ -179,6 +179,7 @@ destroy_remote:  ## Destroy the environment
 		fqdn=`jq -r '.tsb_fqdn' terraform.tfvars.json`; \
 		address=`jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" outputs/terraform_outputs/terraform-tsb-mp.json`; \
 		cd "tsb/fqdn/$$cloud"; \
+		terraform init; \
```
nacx, line 221: Why is the `init` needed on a destroy phase?

Makefile
```diff
@@ -201,7 +202,10 @@ destroy_%:
 		echo "cloud=$* region=$$region cluster_id=$$index"; \
 		cd "infra/$*"; \
 		terraform workspace select $*-$$index-$$region; \
-		terraform destroy ${terraform_destroy_args} -var-file="../../terraform.tfvars.json" -var=$*_k8s_region=$$region -var=cluster_id=$$index; \
+		cluster_name=`terraform output cluster_name | jq . -r`; \
+		terraform state rm "module.external_dns.helm_release.external_dns[0]"; \
+		terraform state rm "module.external_dns.module.external_dns_helm.helm_release.this[0]"; \
+		terraform destroy ${terraform_destroy_args} -var-file="../../terraform.tfvars.json" -var=$*_k8s_region=$$region -var=cluster_id=$$index -var=cluster_name=$$cluster_name; \
```
nacx, line 246: Can you explain why the external-dns needs to be removed manually in this step?

modules/addons/aws/external-dns/main.tf
```diff
@@ -0,0 +1,61 @@
...
+  helm_create_namespace   = true
+  namespace               = "external-dns"
+  helm_timeout            = "900"
+  helm_wait               = true
+
+  values = yamlencode({
+    "policy": "upsert-only"
+    "registry" : "txt"
+    "txtOwnerId" : var.cluster_name
+    "domainFilters": [try(aws_route53_zone.cluster[0].name,"")]
```
nacx: These filters should always be there probably, otherwise we get unexpected random domain names registered an will pollute the DNS.
Let's have it fail if the cluster name can't be read instead of silently defaulting to no domain filters.

modules/addons/aws/external-dns/main.tf
```diff
@@ -0,0 +1,61 @@
+resource "aws_route53_zone" "cluster" {
+  count  = var.external_dns_enabled == true ? 1 : 0
+  name   = "${var.cluster_name}.${var.dns_zone}"
+  tags   = merge(var.tags, {
+          Name = "${var.cluster_name}.${var.dns_zone}"
+  })
+}
```
nacx, line 6: This will create the zone according to the `dns_zone` configured in the variables, but the two resources below only take into account when that zone is in the "shared zone".

Now teams have their own accounts. For example, in the MP team, we've been assigned the `mp.aws.tetrate.com` zone in the MP AWS account. With the current code, the cluster subzone that will be created there will not work.

I think we should get rid of the distinction between shared/private, and do all this stuff always, assuming that each account will already have a properly configured DNS zone, which is where we're going.

modules/aws/jumpbox/aws_cleanup.sh.tmpl
```diff
@@ -3,6 +3,34 @@
...
+    record_sets=$(aws route53 list-resource-record-sets --hosted-zone-id $zone_id --query "ResourceRecordSets[?Type != 'NS' && Type != 'SOA']")
+
+    if [ -z "$record_sets" ]; then
+      echo "No DNS records found in hosted zone with ID $zone_id"
+    else
+      aws route53 change-resource-record-sets --hosted-zone-id $zone_id --change-batch "{\"Changes\": $(echo $record_sets | jq "{ \"Action\": \"DELETE\", \"ResourceRecordSet\": . }")}"
+      echo "Deleted DNS records in hosted zone with ID $zone_id and name $zone_name"
+    fi
+
+    aws route53 delete-hosted-zone --id $zone_id
```
nacx: Shouldn't we defer to terraform and not the script the deletion of the zone it created?

modules/azure/k8s/outputs.tf
```diff
@@ -32,9 +32,13 @@ resource "local_file" "kubeconfig" {
...
+}
+
+output "kubelet_identity" {
+  value = azurerm_kubernetes_cluster.k8s.kubelet_identity
 }
 
 output "locality_region" {
   value = var.location
-}
+}
```
nacx, line 44: I see no script to cleanup records created by external-dns in Azure. Are they cleaned up?

modules/aws/jumpbox/aws_cleanup.sh.tmpl
nacx: Also, if users are creating records inside the `cluster_zone` (and we can have external-dns only watch for those cluster-zone records), then just deleting the zone via terraform would be enough, correct?

The rest will be handled by the register_fqdn, as all other records should be scoped to the cluster zone and then this script wouldn't be neeed.

modules/gcp/jumpbox/gcp-cleanup.sh
```diff
@@ -3,4 +3,25 @@
 export PROJECT=$1
 
 echo 'Destroying K8s Stale PVCs...'
-for pvc in $(gcloud compute disks list --project $PROJECT --filter="name:pvc*" --uri);do echo "Removing $pvc..."; gcloud compute disks delete $pvc --quiet; done
\ No newline at end of file
+for pvc in $(gcloud compute disks list --project $PROJECT --filter="name:pvc*" --uri);do echo "Removing $pvc..."; gcloud compute disks delete $pvc --quiet; done
+
+# Get a list of all managed zones in the project
+ZONES=$(gcloud dns managed-zones list --project="$PROJECT" --format="value(name)")
```
nacx: Do we really need this?
If this about destroying the cluster zones, why simply destroying the zone via terraform is not enough?

Makefile
gonzaloserrano: this bit me last week, when doing destroy it complained of a "missing register_fqdn module" or something similar. Adding this init fixed it.

gonzaloserrano:
```
$ make destroy
╷
│ Error: Module not installed
│
│   on main.tf line 8:
│    8: module "register_fqdn" {
│
│ This module is not yet installed. Run "terraform init" to install all modules required by this configuration.
╵
make: *** [destroy_remote] Error 1
```

Makefile
smarunich: I will move the external-dns to the separate execution runtime, like we have argocd and other addons, otherwise it is a chicken and egg problem, where the k8s cluster can be destroyed prior helm chart deletion.

modules/addons/aws/external-dns/main.tf
smarunich: sounds good.

modules/aws/jumpbox/aws_cleanup.sh.tmpl
smarunich: unfortunately, only azure is able to tolerate the zone deletion with the existing records, however GCP and AWS deletion will fail, if there are any DNS records defined... as DNS records managed out of terraform, we have to have a similar hack :| for clean up applied.

When zone will be removed by the script, terraform will be able to discover the deletion.

modules/azure/k8s/outputs.tf
smarunich: yes, azure apis do allow zone termination with the existing records.

modules/gcp/jumpbox/gcp-cleanup.sh
smarunich: GCP APIs don't allow the zone deletion with the existing DNS records.

smarunich:
Updated to have `external-dns` addon respect the rest of addons framework, like for example `argocd`, but `external-dns` requires proper destroy as modules do orchestrate a cloud-based infrastructure elements, so added a `destroy_external_dns` target as well.

addons/aws/external-dns/variables.tf
```diff
@@ -0,0 +1,74 @@
...
+variable "external_dns_sources" {
+  default = "service"
+}
+
+variable "external_dns_interval" {
+  default = "5s"
+}
+
+variable "external_dns_aws_dns_zone" {
+  default = "aws.sandbox.tetrate.io"
```
nacx: Let's make this a mandatory parameter. Effectively next week the shared accounts are no longer allowed and we should have users explicitly configure that, or take it from the terraform status, but we can't set a default here.

addons/aws/external-dns/variables.tf
```diff
@@ -0,0 +1,74 @@
...
+locals {
+  k8s_regions = var.aws_k8s_regions
+}
+
+variable "tetrate_owner" {
+  default = null
+}
+variable "tetrate_team" {
+  default = "tetrate-service-bridge-sandbox"
+}
```
nacx: These two should be mandatory and not provided by default. We need to enforce proper ownership tagging. Everyone should configure their team and owner name.

addons/azure/external-dns/variables.tf
```diff
@@ -0,0 +1,74 @@
...
+locals {
+  k8s_regions = var.azure_k8s_regions
+}
+
+variable "tetrate_owner" {
+  default = null
+}
+variable "tetrate_team" {
+  default = "tetrate-service-bridge-sandbox"
+}
```
nacx: Same here. Let's just remove these defaults.

addons/gcp/external-dns/variables.tf
```diff
@@ -0,0 +1,73 @@
...
+locals {
+  k8s_regions = var.gcp_k8s_regions
+}
+
+variable "tetrate_owner" {
+    default = null
+}
+variable "tetrate_team" {
+    default = "tetrate-service-bridge-sandbox"
+}
```
nacx: As with other providers, remove the defaults

modules/addons/aws/external-dns/external-dns_aws_cleanup.sh.tmpl
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
```
nacx, line 4: Should we `set -e` to make sure this returns early with a non-zero code if something goes wrong?

modules/aws/k8s/main.tf
```diff
@@ -1,3 +1,6 @@
+
+
+
```
nacx:
```suggestion
```

nacx (commented):
Added some final comments. Overall looks good.

Another thing to consider is adding a note in the `addons/README` explaining that when using the `external-dns` addon you must include the clustername as the DNS prefix when annotating k8s services, otherwise it will not work.

modules/addons/aws/external-dns/external-dns_aws_cleanup.sh.tmpl
smarunich: agree, resolved.

addons/gcp/external-dns/variables.tf
smarunich: agree, resolved.

addons/aws/external-dns/variables.tf
smarunich: agree, resolved.

nacx (approved):
Nice!

PR #251, opened 2023-03-09
Fail early if there is any error

nacx:
This configures the Makefile to fail early if an error occurs.

The current setup uses `;` to concatenate the commands to run, but that ignores failures and continues, using the last command exit code for the overall operation. That last command is usually a `cd ..`, which results in always succeeding, and the provisioning process is not aborted if any previously required step failed.

This is quite annoying as you don't know when things went wrong, especially when destroying stuff, as you may end up removing the local status while leaving garbage there.

This PR forces an exit on the first failure so that the process does not continue when it shouldn't.


smarunich (approved):

smarunich:
oh indeed a very needed one!

PR #252, opened 2023-03-09
Automatically configure a VM endpoint

### LINKED ISSUE #184: Configure mesh expansion with the repository and onboarding plane endpoint

nacx:
Fixes https://github.com/tetrateio/tetrate-service-bridge-sandbox/issues/184

Now that we can install external-dns we can automatically configure the VM endpoint.

### DIFF
--- a/modules/tsb/cp/main.tf
+++ b/modules/tsb/cp/main.tf
@@ -89,6 +89,7 @@ resource "helm_release" "controlplane" {
     ratelimit_enabled               = var.ratelimit_enabled
     ratelimit_namespace             = var.ratelimit_namespace
     identity_propagation_enabled    = var.identity_propagation_enabled
+    vm_endpoint                     = "vms.${var.cluster_name}.${regep("[^\\.]*\\.(.*)", var.tsb_fqdn)}"
   })]
 
   set {
--- a/modules/tsb/cp/main.tf
+++ b/modules/tsb/cp/main.tf
@@ -65,6 +65,18 @@ data "local_file" "service_account" {
   depends_on = [null_resource.jumpbox_tctl]
 }
 
+data "kubectl_path_documents" "vm_certs" {
+  pattern = "${path.module}/manifests/cert-manager/vm-onboarding-cert.yaml.tmpl"
+  vars = {
+    vm_endpoint = var.vm_endpoint
+  }
+}
+
+resource "kubectl_manifest" "vm_certs" {
+  count     = length(data.kubectl_path_documents.vm_certs.documents)
+  yaml_body = element(data.kubectl_path_documents.vm_certs.documents, count.index)
+}
+
 resource "helm_release" "controlplane" {
   name                = "controlplane"
   repository          = var.tsb_helm_repository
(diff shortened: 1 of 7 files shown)


modules/tsb/cp/main.tf
```diff
@@ -89,6 +89,7 @@ resource "helm_release" "controlplane" {
     ratelimit_enabled               = var.ratelimit_enabled
     ratelimit_namespace             = var.ratelimit_namespace
     identity_propagation_enabled    = var.identity_propagation_enabled
+    vm_endpoint                     = "vms.${var.cluster_name}.${regep("[^\\.]*\\.(.*)", var.tsb_fqdn)}"
```
nacx: the regexp is basically removing the first block to capture the DNS suffix

modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
```diff
@@ -78,7 +87,13 @@ spec:
...
           uri: redis-master.${ratelimit_namespace}.svc.cluster.local:6379
     %{ endif }
-  meshExpansion: {}
+  meshExpansion:
+    onboarding:
+      localRepository: {}
+      endpoint:
+        hosts:
+          - ${vm_endpoint}
+        secretName: vm-onboarding
```
nacx, line 96: Opened as draft to see if it is OK to configure this via ext-dns. If the approach is OK I'll add the bots to generate the certificate and populate this secret, which is still not in this PR.

modules/tsb/cp/main.tf
smarunich: to keep it generic... should we have a vm_endpoint as a variable for tsb_cp module ... $var.tsb_fqdn may be or can be different from external-dns domain... so basically under module itself... we should define vm_endpoint as

vm_endpoint = $var.cluster_name + lookup/element to select external_dns_$cloud_dns_zone

modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
smarunich: oh yeah, so heavily looking forward to the trusted certificate :)

tsb/cp/variables.tf
```diff
@@ -139,4 +139,8 @@ variable "ratelimit_enabled" {
 
 variable "identity_propagation_enabled" {
   default = false
-}
\ No newline at end of file
+}
+
+variable "vm_endpoint" {
+  default = "vms.${var.cluster_name}.${regep("[^\\.]*\\.(.*)", var.tsb_fqdn)}"
```
smarunich: regep > `regexp`

PR #269, opened 2023-04-20
Removed set -e on destroy

shamusx:
As we are adding more features for cleanup - I am finding if we don't do the full tsb deploy, or if state is partially cleaned, and we do a `make destroy`, the destroy will fail as it's trying to find static file points based on variables file.

I think for destroy we can remove the `set -e`


smarunich (approved):
yes, the destroy process has to be improved, on top if the automation is only used to standup k8s, but no tsb - the destroy will fail as well, so let's open an issues when destroy fails to improve the process...

PR #278, opened 2023-06-28
Small helper to run remote commands

nacx:


smarunich (commented):
sorry on my phone today, how it should work? does it impact the existing ssh (no command) run?

nacx:
It doesn't change if you just run the `./ssh`; it will connect you to the server as before. With this you can also do `./ssh tctl foo bar` or run any command in the jumbox without having to enter it

smarunich (approved):

boeboe:
Have you considered this TF provider in case `null-provider` or `remote-exec` wasn't _elegant_ enough?


https://registry.terraform.io/providers/loafoe/ssh/latest/docs

Might help in clean-up/tear-down as well (eg aws eks LB destruction).

PR #297, opened 2023-09-15
Update controlplane-values.yaml.tmpl

shamusx:
enable internalWasmExtension

### DIFF
--- a/modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
+++ b/modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
@@ -26,6 +26,8 @@ spec:
     clusterName: ${cluster_name}
     selfSigned: true
   components:
+    istio:
+      mountInternalWasmExtensions: true
     xcp:
       centralAuthMode: JWT
       configProtection: {}


modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
```diff
@@ -26,6 +26,8 @@ spec:
     clusterName: ${cluster_name}
     selfSigned: true
   components:
+    istio:
+      mountInternalWasmExtensions: true
```
nacx, line 30: Nice! Keep in mind that this only works in TSB >= 1.7.

nacx (approved):

modules/tsb/cp/manifests/tsb/controlplane-values.yaml.tmpl
nacx: And also keep in mind that if there are VMs in that cluster, they need the onboarding-agent package upgraded to the version that comes with 1.7 as well.

shamusx: thank you!

PR #298, opened 2023-09-25
[ON_HOLD] enable `enableWasmDownloadProxy` by default for 1.8

smarunich:


nacx:
You may want to hold on enabling this until https://github.com/tetrateio/tetrate/issues/17828 is fixed... It is a known issue in 1.7.0

PR #300, opened 2023-10-11
Refactor TFVAR system - FOR DISCUSSION ONLY

boeboe:
**Small summary**

- updated docs according to below
- refactored tfvar.json schema
- make tfvar.json a makefile param that can be overriden
- moved to make.sh as inline makefile bash was not sufficient
- added more debug capabilities (print statement) => potentially hide them behind a DEBUG=false/true flag
- added schema validation as makefile target
- added json schema for VSCode or other IDEs
- added schema transformation script to migrate from old to new json
- some bug fixes I bumped on during refactoring (AWS cleanup, dns_provider extrapolation based on tsb fqdn,...)
- started to clean up on TF: 
    - all variables should be properly documented, typed and checked
    - introduced a new way to enforce mandatory vs optional params + default values (check tags and tsb structs)
    - `terraform fmt` is a must for mental sanity
    - cleanup copy paste variables.tf data from unneeded params (makes debug really hard)
- TODO (after this crazy PR is processed)
    - more code cleanup
    - more docs and parameter cleanup
    - generate TF docs per module (readme, html, whatever) and commit to git
 - extend on per cluster configuration
    - (prio1) add zones as optional to scope clusters (Box.com as use case)
    - other common k8s tweaks


smarunich:
let's start simple, make.sh need to include the existing Makefile targets - let's not drop the existing functionality: https://github.com/tetrateio/tetrate-service-bridge-sandbox/blob/main/Makefile - for example build/destroy specific cloud, that is a common use case, or onboard specific cloud only to TSB. 

I feel like we need to chunk this PR to few when merging to monorepo, as the size of it... and ability to catch all drops in/outs needs time to converge on, by any chance is there a way to split the changes across few PRs? For example, move tflint to a different PR.

Before merging to main as we converge on the common format, one of us have to converge update vars to k8s_clusters to reflect https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/terraform.tfvars.json.sample

PR #301, opened 2023-10-12
fix aws cleanup and dns_provider extrapolation based on tsb fqdn

boeboe:
1. The AWS cleanup script sometimes failed pre-maturely because it also tried to delete the default SG associated with the VPC, which is not allowed (per AWS design). Since we used `set -e`, the script would stop and potentially not delete the EKS related SG's, leading to failure of VPC cleanup.

2. The logic to extrapolate dns_provider based on TSB FQDN did not match in case AWS was used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ tsb_mp:  ## Deploys MP
 		set -e; \
 		cloud=`jq -r '.tsb_mp.cloud' terraform.tfvars.json`; \
 		dns_provider=`jq -r '.dns_provider' terraform.tfvars.json`; \
-		[ "$$dns_provider" == "null" ] && dns_provider=`jq -r '.tsb.fqdn' terraform.tfvars.json | jq -Rr 'split(".")[1] | if . == "azure" then "azure" elif . == "gcp" then "gcp" else "aws" end'`; \
+		[ "$$dns_provider" == "null" ] && dns_provider=`jq -r '.tsb_fqdn' terraform.tfvars.json | cut -d"." -f2 | sed 's/sandbox/gcp/g'`; \
 		cd "tsb/mp"; \
 		terraform workspace select default; \
 		terraform init; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -118,7 +118,7 @@ tsb_mp:  ## Deploys MP
 		set -e; \
 		cloud=`jq -r '.tsb_mp.cloud' terraform.tfvars.json`; \
 		dns_provider=`jq -r '.dns_provider' terraform.tfvars.json`; \
-		[ "$$dns_provider" == "null" ] && dns_provider=`jq -r '.tsb_fqdn' terraform.tfvars.json | cut -d"." -f2 | sed 's/sandbox/gcp/g'`; \
+		[ "$$dns_provider" == "null" ] && dns_provider=`jq -r '.tsb.fqdn' terraform.tfvars.json | jq -Rr 'split(".")[1] | if . == "azure" then "azure" elif . == "gcp" then "gcp" else "aws" end'`; \
```
smarunich: at this stage the most common domain that is being used is "sandbox.tetrate.io" that is behind gcp and does not include cloud prefix azure, gcp or custom aws-ce/se.. etc domain so keeping it as is as technically this interpolation, also have tested provided interpolation update has a typo.

smarunich (approved):

Makefile
boeboe: Can we update the Readme.md on this and have an inline comment? I wasn't aware there was an fqdn without a reference to a cloud provider. Also, we should be able to extrapolate in all cases, to keep consistency. I prefer more sed/jq/cut magic to cover all cases, as opposed to undocumented assumptions on popular input patterns.

boeboe: As an alternative... drop this line all together and make `dns_provider` a hard mandatory parameter with fail fast feedback.

smarunich: yeah fair point on documenting as we migrate this to help shell function, dns_provider was a mandatory before, but we dropped to "keep it simple" terraform.tfvars.json file for the user simplicity sake, as you don't want to overload user with fail fast either.

PR #302, opened 2023-10-13
Propose new structure make/bash and general framework

smarunich:


Makefile
```diff
@@ -21,6 +24,11 @@ help: Makefile ## This help
 init:  ## Terraform init
 	@echo "Please refer to the latest instructions and terraform.tfvars.json file format at https://github.com/tetrateio/tetrate-service-bridge-sandbox#usage"
 
+
+.PHONY: validate
+validate:  ## Validate terraform input
+	export TFVARS_JSON=${tfvars_json} && ./make/validate.sh validate_input
```
smarunich: should we make validate required and be part if variables.sh and basically be mandatory for each run, we can have it as a separate make target too, but by default it should run everytime in my opinion as variables.sh is executed.

make/addon.sh
```diff
@@ -0,0 +1,100 @@
...
+fi
+
+# Check if the file pointed to by TFVARS_JSON does not exist and exit.
+if [ ! -f "${TFVARS_JSON}" ]; then
+  print_error "File '${TFVARS_JSON}' does not exist."
+  exit 1
+fi
+
+# Validate input values.
+if [[ "${ACTION}" == "deploy_addon" || "${ACTION}" == "destroy_addon" ]]; then
```
smarunich: let's avoid complex IFs for the future maintainability in favour of arrays like

```
CLOUD_PROVIDER="aws"
CLOUD_PROVIDERS=("aws" "azure" "gcp")

# Check if "a" is in the array
if [[ ! " ${CLOUD_PROVIDERS[*]} " == *" $CLOUD_PROVIDER "* ]]; then
    print_error "Invalid cloud provider. Must be one of 'azure', 'aws', or 'gcp'."
fi
```

make/addon.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
+#
+# Helper script for addons: deploy and destroy of argocd, fluxcd, tsb-monitoring and external-dns.
+
+export BASE_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source ${BASE_DIR}/helpers.sh
```
smarunich: variables.sh with variable validation to be enabled by default

make/helpers.sh
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bash
```
smarunich, line 1: need to include variable initialization and validation, the rest of functions to refer as it is already technically done... just general comment not waste paper, but to keep it as a reminder.

make/k8s.sh
```diff
@@ -0,0 +1,104 @@
...
+fi
+
+# Check if the file pointed to by TFVARS_JSON does not exist and exit.
+if [ ! -f "${TFVARS_JSON}" ]; then
+  print_error "File '${TFVARS_JSON}' does not exist."
+  exit 1
+fi
+
+# Validate input values.
+if [[ "${ACTION}" == "deploy_k8s" || "${ACTION}" == "destroy_k8s" || "${ACTION}" == "refresh_token_k8s" ]]; then
```
smarunich: need to use arrays as previously mentioned

make/k8s.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
```
smarunich: technically this is not k8s only, it is infra_cloud module, naming convention should be 

infra_aws for example, k8s_auth function should be separate of this in my opinion and referenced by modules that do use kubernetes provider, please refer to https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/make/infra_aws.sh and https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/make/tetrate_managementplane.sh for example

make/tsb.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
smarunich, line 1: let's re-use as much as we can of https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/make/tetrate_managementplane.sh and have a separate helper for controlplane https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/make/tetrate_controlplane.sh

make/validate.sh
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env bash
```
smarunich: Reminder to provide on error the suggested output and existing variable (if set), basically introduce "struct" as @boeboe suggested before to have a proper variables management.

terraform.tfvars.json.sample
```diff
@@ -1,25 +1,24 @@
...
-    "tsb_image_sync_apikey": "<TSB_REPO_APIKEY>",
-    "tsb_password": "Tetrate123",
-    "tsb_mp": {
-        "cloud": "azure",
-        "cluster_id": 0
-    },
-    "tsb_org": "tetrate",
-    "aws_k8s_regions": [
-    ],
+    "aws_k8s_regions": [],
```
smarunich, line 2: need to align to something like 
```
{
    "name_prefix": "<YOUR UNIQUE PREFIX NAME TO BE CREATED>",
    "tetrate": {
        "version": "1.7.0+tse",
        "image_sync_username": "<TETRATE_REPO_USERNAME>",
        "image_sync_apikey": "<TETRATE_REPO_APIKEY>"
    },
    "k8s_clusters": {
        "aws": [
        {
            "name": "demo1",
            "region": "us-west-1",
            "version": "1.24",
            "management_plane": "true", // only one cluster can be management plane
            "tier1": "true",  
            "external_dns_zone": "aws-ce.sandbox.tetrate.io"
        },
```
https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/terraform.tfvars.json.advanced.sample

smarunich (commented):
@boeboe @shamusx

smarunich:
to keep the effort simple can you land variables.sh (helper.shz) https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/make/variables.sh and https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/make/infra_aws.sh based on the combined feedback with the update to Makefile so we can agree and eventually scale it up to the rest of functions

Makefile
boeboe: I my avalanche branch, I had it as a prerequisite target on the deploy targets. forcing it to run automatically, yet having the flexibility to run it separately as well (to debug a broken json).

make/addon.sh
boeboe: Ack, adopted

make/k8s.sh
boeboe: ack, adopted

make/k8s.sh
boeboe: ack, adopted to some extent.

terraform.tfvars.json.sample
boeboe: This will be done later. First move over to shell scripts.

boeboe:
New changes:

- adopted code review changes
- started using linting tool to avoid bash shenanigans
   - https://www.shellcheck.net
   - https://marketplace.visualstudio.com/items?itemName=timonwong.shellcheck
- added parameter input validation on bash functions to increase debug capabilities

Next step: migrate jq => bash functions.

make/validate.sh
```diff
@@ -0,0 +1,148 @@
...
+  [[ -z "${1}" ]] && print_error "Please provide tfvar.json file as 1st argument" && return 1 || local json_tfvars="${1}" ;
+
+  # Check if the file exists
+  if [[ ! -f "${json_tfvars}" ]]; then
+    print_error "File '${json_tfvars}' does not exist."
+    exit 1
+  fi
+
+  # Validate 'aws_k8s_regions'
+  if ! jq -e '.aws_k8s_regions' "${json_tfvars}" > /dev/null; then
```
smarunich: wdyt if we can avoid repetitive ifs with something like, we also need to list each key full path, i.e. tsb_mp.cloud, etc etc basically to keep it generic

```bash

export json_tfvars="terraform.tfvars.json"

export REQUIRED_VARS=("dns_provider:aws,azure,gcp" "tsb_mp.cluster_id:" "tetrate_owner:")

for item in "${REQUIRED_VARS[@]}"; do
    option="${item%%:*}";  # Extracts everything before the colon
    allowed_values="${item##*:}"; # Extracts everything after the colon
    current_value=$(jq -r ".$option" ${json_tfvars})
    if [[ -z "$current_value" ]]; then
        echo "Missing $option in the JSON.";
    fi 
    if ! [[ -z "$allowed_values" ]] && ! [[ -z $(grep -q $allowed_values <<< $current_value) ]] ; then
        echo "$option is set to the incorrect value: $current_value, allowed values: $allowed_values \n";
    fi
done

```

make/variables.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
smarunich, line 1: as discussed `validate_input` should be invoked with variables.sh

terraform.tfvars.json.sample
```diff
@@ -1,25 +1,24 @@
...
     "azure_k8s_regions": [
         "eastus"
     ],
+    "dns_provider": "azure",
```
smarunich, line 6: this is an optional property, let's not promote it to required.

make/addon.sh
```diff
@@ -0,0 +1,94 @@
...
+source "${BASE_DIR}/helpers.sh"
+# shellcheck source=/dev/null
+source "${BASE_DIR}/variables.sh"
+
+ACTION=${1}
+CLOUD_PROVIDER=${2}
+ADDON_NAME=${3}
+
+# Validate input values.
+SUPPORTED_ACTIONS=("deploy_infra" "destroy_infra" "refresh_token_k8s")
```
smarunich: thinking out loud without ability to test it, I assume supported_actions will reflect makefile targets?

make/variables.sh
smarunich: to be fair validate probably should reside with variables.sh... thinking out loud

make/validate.sh
boeboe: Done

make/variables.sh
boeboe: I have moved it to `variables.sh`

The disadvantage was that you could no longer invoke it as a separate `makefile` target for pre-run validation without the intent to start deploying, but mitigated that by invoking `variables.sh` in the `init` target.

terraform.tfvars.json.sample
boeboe: Check :heavy_check_mark:

make/addon.sh
boeboe: I will align `SUPPORTED_ACTIONS` to match the exact same name as the original `makefile` targets.

make/validate.sh
boeboe: Done :heavy_check_mark:

PR #307, opened 2023-10-18
make2bash: external-dns deploy/destroy update

smarunich:
make2bash: external-dns deploy/destroy update

Only deploy if required variable is set, we should scale this pattern to the rest of addons, e.g. if EXTERNAL_DNS_AWS_DNS_ZONE is set, external-dns for aws will get deployed, otherwise the respective error for missing variable will be provided

### DIFF
--- a/make/addon.sh
+++ b/make/addon.sh
@@ -74,6 +74,7 @@ function deploy_addon_per_region() {
   set -e
 
   local index=0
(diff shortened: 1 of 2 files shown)


shamusx (approved):

boeboe (approved):
LGTM

PR #309, opened 2023-10-26
Introduce new JSON schema

boeboe:
- New JSON schema
- Examples migrated
- New helper functions
- Refactored `jq` magic and iterations in the shell scripts
- Pass on `cluster` and `addon_config` objects as terraform `-var=...` iso x amount of vars
- Add support for Argo FQDN exposure
- Deprecated Tier1 support
- Added a migration script to new JSON schema
- Kept DNS provider support


smarunich:
not suggesting that this is an ideal approach: https://github.com/tetrateio/tetrate-service-bridge-sandbox/pull/309/commits/0d9cecaaa459a8390955698ec74cccf8628797bb#diff-a86cc0bb6aa6cd0d6d4ed8384e11316a90d232f310ed9cfa0d7100032b7b2265, i.e. I had to apply in order to unblock the testing...

however having `local.tetrate_defaults` is a map to have a parity with var.tetrate as map not as object type, does simplify the merge between, however object type does provide a better representation of how the object should be defined, I hit the same dilemma when crafting tse sandbox. Basically having an open map as any gives an ability to instantiate empty variable that can be after on expanded with defaults.

Technically map can be involved to something like to have variable constraints defined as well as per chatgpt...

```
variable "tetrate" {
  description = "An object containing global tetrate configuration"
  type        = map(any)
  default     = {}

  validation {
    condition     = can(regex("^.+\\..+\\..+$", var.tetrate.fqdn))
    error_message = "The fqdn must be a valid fully qualified domain name."
  }

  validation {
    condition     = length(var.tetrate.password) >= 8
    error_message = "The password must be at least 8 characters long."
  }

  // Add additional validations as necessary
}
```

smarunich:
as of now based on the applied changes no issues with `make k8s` step, hitting the below when applying `make tsb` will continue tomorrow... it has to be "path" issue as by itself error is not a significant constraint.

```
│ Error: Insufficient features blocks
│
│   on <empty> line 0:
│   (source code not available)
│
│ At least 1 "features" blocks are required.
╵
make: *** [tsb_mp] Error 1
```

vars file

```
{
  "k8s_clusters": {
    "aws": [],
    "azure": [
      {
        "region": "eastus",
        "tetrate": {
          "control_plane": true,
          "management_plane": true
        }
      },
      {
        "region": "eastus",
        "tetrate": {
          "control_plane": true
        }
      }
    ],
    "gcp": []
  },
  "tags": {
    "tetrate_owner": "sergey",
    "tetrate_team": "ce"
  },
  "name_prefix": "r172dev7",
  "tetrate": {
    "fqdn": "r172dev7.azure.sandbox.tetrate.io",
    "image_sync_apikey": "<REDACTED>",
    "image_sync_username": "<REDACTED>",
    "organization": "tetrate",
    "password": "<REDACTED>",
    "version": "1.7.2"
  }
}
```

make/tsb.sh
```diff
@@ -36,43 +36,50 @@ function help() {
...
-  local fqdn=$(jq -r '.tsb_fqdn' "${TFVARS_JSON}")
+  local cluster=$(get_mp_cluster_config "${TFVARS_JSON}")
+  local fqdn=${TETRATE_FQDN}
   local address=$(jq -r "if .ingress_ip.value != \"\" then .ingress_ip.value else .ingress_hostname.value end" outputs/terraform_outputs/terraform-tsb-mp.json)
-  local dns_provider=$(jq -r '.dns_provider' "${TFVARS_JSON}")
-  if [ "${dns_provider}" == "null" ]; then
-    dns_provider=$(jq -r '.tsb_fqdn' "${TFVARS_JSON}" | cut -d"." -f2 | sed 's/sandbox/gcp/g')
-  fi
-  print_info "Going to deploy tsb management plane fqdn on cloud '${dns_provider}'"
+  local dns_provider=$(get_cluster_cloud "${cluster}")
```
shamusx: have some scenarios with dns_provider will be different then fqdn provider, so will work to add logic similar to:
```sh
  local dns_provider=$(jq -r '.dns_provider' "${TFVARS_JSON}")
  if [ "${dns_provider}" == "null" ]; then
    dns_provider=$(jq -r '.tsb_fqdn' "${TFVARS_JSON}" | cut -d"." -f2 | sed 's/sandbox/gcp/g')
  fi
```
to variables.sh

smarunich:
Resolved conflicts, have been independently reviewing with @shamusx , need tomorrow to do manual testing and we should be good to merge

smarunich (approved):

shamusx (approved):

PR #317, opened 2023-11-06
ES upgrade to 8.10.4 + Moving back to built-in cert-manager with AUTO mode for control planes, leveraging built-in TSB capabilties to generate Istiod cacerts

smarunich:
1. Upgrade ES to 8.10.4
2. Setting cert-manager to AUTO for control planes
3. Leveraging TSB to generate and distribute Istiod cacerts

### DIFF
--- a/modules/addons/elastic/manifests/es.yaml
+++ b/modules/addons/elastic/manifests/es.yaml
@@ -5,7 +5,7 @@ metadata:
   name: tsb
   namespace: elastic-system
 spec:
-  version: 7.13.2
+  version: 8.10.4
   http:
     service:
       spec:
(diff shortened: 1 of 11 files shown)


modules/addons/elastic/manifests/es.yaml
```diff
@@ -36,7 +37,7 @@ metadata:
   name: tsb
   namespace: elastic-system
 spec:
-  version: 7.13.2
+  version: 8.10.4
```
shamusx, line 40: Should make this one a variable, and same for the helm install of operator similar to cert-manager.

smarunich:
closed in favour to https://github.com/tetrateio/tetrate-service-bridge-sandbox/pull/326

PR #318, opened 2023-11-06
`external-dns` variable check `if empty` fix typo

smarunich:
external-dns variable check if empty fix typo


smarunich:
to be super-seeded by https://github.com/tetrateio/tetrate-service-bridge-sandbox/pull/309

PR #333, opened 2023-11-10
modified tsb.sh to support aws-ce.sandbox.tetrate.io

AlexeySkvortsov:
updated `tsb.sh` to support EKS MP and AWS dns public hosted zone `aws-ce.sandbox.tetrate.io`

### DIFF
--- a/make/tsb.sh
+++ b/make/tsb.sh
@@ -78,11 +78,7 @@ function deploy_mp_fqdn() {
   echo "Processing cluster:" 
   echo "${cluster}" | jq '.'
 
-  if [[ "${dns_provider}" == aws* ]]; then
-    run "pushd tsb/fqdn/aws"
-  else
-    run "pushd tsb/fqdn/${dns_provider} > /dev/null"
-  fi
+  run "pushd tsb/fqdn/${dns_provider} > /dev/null"
   run "terraform workspace new ${workspace} || true"
   run "terraform workspace select ${workspace}"
   run "terraform init"
--- a/make/tsb.sh
+++ b/make/tsb.sh
@@ -78,7 +78,11 @@ function deploy_mp_fqdn() {
   echo "Processing cluster:" 
   echo "${cluster}" | jq '.'
 
-  run "pushd tsb/fqdn/${dns_provider} > /dev/null"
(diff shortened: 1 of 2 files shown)


make/tsb.sh
```diff
@@ -78,7 +78,11 @@ function deploy_mp_fqdn() {
   echo "Processing cluster:" 
   echo "${cluster}" | jq '.'
 
-  run "pushd tsb/fqdn/${dns_provider} > /dev/null"
+  if [[ "${dns_provider}" == aws* ]]; then
```
shamusx: Hmm if you provide `dns_provider` in variables file this shouldn't happen.
`tetrate.dns_provider`  I almost wonder if we should make dns_provider required, or our `variables.sh` should narrow this down more effectively not the `tsb.sh`

AlexeySkvortsov:
@shamusx agreed. moved to variables.sh (also to avoid double set for `deploy` and `destroy`)

make/variables.sh
```diff
@@ -80,6 +80,9 @@ validate_input "${TFVARS_JSON}"
 # Parse tfvars.json and export variables
 # DNS Provider variable
 export DNS_PROVIDER=$(jq -r '.tetrate.dns_provider // .tetrate.fqdn | select(type == "string") | split(".") | if length > 1 then .[1] else .[0] end | select(. != null) | sub("sandbox"; "gcp")' "${TFVARS_JSON}")
+if [[ "${DNS_PROVIDER}" == aws* ]]; then
```
shamusx: If we update line 82 to looks for specific prefix instead like:
`jq -r '.tetrate.dns_provider // .tetrate.fqdn | select(type == "string") | split(".") | if length > 1 then .[1] else .[0] end | select(. != null) | sub("sandbox"; "gcp") | capture("^(?<cloud>aws|gcp|azure)") | .cloud'`

Would this achieve what your aiming for?  this way we can account for any cloud being a prefix ie: gcp-ce 
WDYT @AlexeySkvortsov

AlexeySkvortsov: @shamusx yes, I like this idea and yes, this achieve what I am aiming for and covers all other cloud prefixes. done.

shamusx (approved):

PR #339, opened 2023-11-27
remove sa onboarding step

shamusx:
1.7.x now defaults the SA org binding

### DIFF
--- a/modules/tsb/cp/manifests/tctl/tctl-controlplane.sh.tmpl
+++ b/modules/tsb/cp/manifests/tctl/tctl-controlplane.sh.tmpl
@@ -4,27 +4,3 @@
 /usr/bin/env tctl config profiles set-current "default"
 /usr/bin/env tctl apply -f ${cluster_name}-cluster.yaml
 /usr/bin/env tctl install cluster-service-account --cluster ${cluster_name} > ${cluster_name}-service-account.jwk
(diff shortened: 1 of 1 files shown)


smarunich (approved):

smarunich (approved):

modules/tsb/cp/manifests/tctl/tctl-controlplane.sh.tmpl
```diff
@@ -4,27 +4,3 @@
 /usr/bin/env tctl config profiles set-current "default"
 /usr/bin/env tctl apply -f ${cluster_name}-cluster.yaml
 /usr/bin/env tctl install cluster-service-account --cluster ${cluster_name} > ${cluster_name}-service-account.jwk
-/usr/bin/env tctl x gitops grant ${cluster_name}
```
sreeharikmarar, line 7: I don't think we have deprecated `tctl x gitops` for `1.8` yet.

smarunich: I don't think we should completely, as we have options where workspace access bindings are being used, so technically tctl x gitops should always stay experimental. But as far as I understand at this stage we generate org-wide `admin` binding by default starting 1.7.x so it should not affect 1.7 or 1.8 branches.

PR #347, opened 2024-01-17
edge/app gw updates

shamusx:
- update names to edge and app for gateways
- remove QA revisioned deployments/configurations
- remove qa IB from CP values


smarunich:
perhaps to rename files too... i.e. tier1 and ingress gateways are in the filenames too: https://github.com/tetrateio/tetrate-service-bridge-sandbox/pull/347/files#diff-8c4f919bc15b6e5939ff3c1599a82d771eebbb5513efc61c9b795df0a43769b8

shamusx:
good point, updated.

PR #349, opened 2024-01-31
During destroy skip workspaces that don't exist instead of failing

morepork:
This means the destroy can be resumed after deleting a workspace if there are still resources in the other workspaces. Previously it would error out and not delete any of the other workspaces.


smarunich (approved):

smarunich:
Thank you @morepork !

PR #378, opened 2024-04-12
update aws cleanup

smarunich:


smarunich:
imported from https://github.com/tetrateio/tetrate-service-express-sandbox/blob/main/modules/aws/base/main.tf

PR #381, opened 2024-04-29
Moving to a single AZ kubernetes clusters, downsizing clusters and jumpboxes

smarunich:
1. downsized kubernetes nodes to all clusters to 2vCPU/8GB
2. downsized jumpboxes to the smallest generally available instances 1vCPU/1-2GB
3. GKE and AKS clusters deployed in the single AZ, to avoid cross-AZ charges
4. EKS is deployed across minimum number of AZs - 2 AZs, enforced by AWS to minimize cross-AZ charges.

### DIFF
--- a/infra/aws/variables.tf
+++ b/infra/aws/variables.tf
@@ -105,7 +105,7 @@ variable "cidr" {
 
 variable "jumpbox_machine_type" {
   type    = string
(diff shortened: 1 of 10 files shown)


shamusx (approved):
looks good to me

